PR #27, opened 2016-10-18
Adding newrelic module to requirements

giorgiosironi:
This module needs to be used inside the virtualenv to run a wrapper around the WSGI command, in order to instrument the application and gather data.

We could in theory install it when the virtualenv is created and only in selected EC2 instances, but the presence of the code doesn't seem invasive (this will not be run in dev, for example.)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -23,3 +23,4 @@ requests==2.9.1
 unittest-xml-reporting==2.1.0
 python-json-logger==0.1.5
 pyflakes==1.3.0
+newrelic==2.70.0.51


lsh-0:
https://pypi.python.org/pypi/newrelic/2.70.0.51

giorgiosironi:
That's him. Integrating through `newrelic-admin` is easier as there is no modification in the Python code.
There is also a newer version released yesterday, 2.72.0.52, nothing major.

lsh-0:
yeah, I'd like to take a look and see what it does before merging this. I don't like magic

giorgiosironi:
For posterity reading this in the future, this one does not do anything as it's just Python code being downloaded, the stuff is at https://github.com/elifesciences/lax-formula/pull/3

PR #34, opened 2016-11-03
Bugfix dateformat

lsh-0:
the v1 pubdate has no time component or timezone component but during scraping in the bot-lax-adaptor it *is* given a time component but no timezone. 

when an article is INGESTED into lax, a datetime isn't set. when an article is PUBLISHED and it is a v1, the published value from the json is used, which lacks the timezone. the statusdate and versiondates however are not pulled from the xml and are managed by lax, so they do have timezones.

this fix replaces the version pubdate from lax when merging article fragments and storing the result. it also has the sideeffect of correctly enforcing the rule of: 'published' is null if article hasn't been published yet (which fails validation, PR on api-raml is still open).

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1,5 +1,8 @@
 #!/bin/bash
 set -e # everything must succeed.
+
+. download-api-raml.sh
+
 if [ ! -d venv ]; then
     virtualenv --python=`which python2` venv
 fi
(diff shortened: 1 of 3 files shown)


giorgiosironi:
I'm not sure where the correct value (with timezone) is coming from, but I would like to see this in action in `end2end` :+1:

lsh-0:
merged

PR #40, opened 2016-11-17
Bugfix handle bad data

lsh-0:


lsh-0:
leaving this PR open as I need the above in prod

PR #43, opened 2016-11-21
Executing smoke tests after every deploy

giorgiosironi:


giorgiosironi:
@lsh-0 a few more GETs never hurt

smoke_tests.sh
```diff
@@ -2,3 +2,4 @@
 set -ex
 
 [ $(curl --write-out %{http_code} --silent --output /dev/null https://$(hostname)/) == 200 ]
+[ $(curl --write-out %{http_code} --silent --output /dev/null https://$(hostname)/api/v2/articles) == 200 ]
```
giorgiosironi: if you can think of other lightweight URLs that should always be working in any environment and any situation, please add them here

lsh-0: /admin
/explorer
both of these should exist but both load the login page as they are protected

rss feeds should exist:
/rss/articles/poa+vor/last-28-days/

PR #44, opened 2016-11-22
added multiprocessing to bulk ingest.

lsh-0:
takes it from ~50mins to ~11mins. 

this is still rather slow! but I suspect it's postgresql and transaction boundaries now. 

I might even write a lightweight bulk ingest script that bypasses the business rules (it *is* a force ingest anyway)


lsh-0:
I wondered about that. Will deal with this presently then. This PR has broken tests too where writing to stdout is failing or is  getting jumbled with something else so that parsing as json fails.

lsh-0:
so - I'm passing around a 'print_queue' `Queue` object now and have the jobs write their success or errors to it.

it seems to work! feels a bit clunky, but the code isn't too obfuscated and shaves a lot of time off of the backfill. 

will test this with a run of many articles tomorrow morning and make sure nothing funky got through.

lsh-0:
@giorgiosironi, ok I have the backfill script/process obeying me again. two bugs not picked up during testing for some reason. very strange.

PR #46, opened 2016-11-25
strict iso8601 dates

lsh-0:
* utils.json_dumps now formats dates with a terminating 'Z' for utc.

### DIFF
--- a/src/publisher/tests/test_utils.py
+++ b/src/publisher/tests/test_utils.py
@@ -14,6 +14,12 @@ class TestUtils(base.BaseCase):
     def tearDown(self):
         pass
 
+    def test_json_dumps_rfc3339(self):
+        dt = datetime(year=2001, month=1, day=1, hour=23, minute=59, second=59, microsecond=123, tzinfo=pytz.utc)
+        struct = {'dt': dt}
+        expected = '{"dt": "2001-01-01T23:59:59Z"}'
+        self.assertEqual(utils.json_dumps(struct), expected)
+    
     def test_resolve_paths(self):
         tests_dir = join(settings.SRC_DIR, 'publisher', 'tests', 'fixtures')
         cases = [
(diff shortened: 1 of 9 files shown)


src/publisher/tests/test_utils.py
```diff
@@ -14,6 +14,12 @@ def setUp(self):
     def tearDown(self):
         pass
 
+    def test_json_dumps_rfc3339(self):
+        dt = datetime(year=2001, month=1, day=1, hour=23, minute=59, second=59, microsecond=123, tzinfo=pytz.utc)
```
giorgiosironi: would test this also with a different `tzinfo` being specified

giorgiosironi:
`api-raml` now has the validation for this, so I'll upgrade it on this branch and check the result.

giorgiosironi:
Ah no, I can't because this is lax and not bot-lax-adaptor. This is an additional barrier against serving bad data to journal, but we need the change in formatting to take place in bot-lax-adaptor for the article JSON to be valid right?

src/publisher/tests/test_utils.py
lsh-0: good call

lsh-0:
> but we need the change in formatting to take place in bot-lax-adaptor for the article 
> JSON to be valid right?

I don't think so actually. The dates that are being handled are 'published', 'statusDate' and 'versionDate' - all of which are handled by lax. I'll double check the 'published' date but I'm pretty sure lax updates that formatting too

giorgiosironi:
Ah, so they are not in the JSON that bot-lax-adaptor produces at all? What's inside those fields there, some placeholder?

giorgiosironi:
Trying out https://github.com/elifesciences/bot-lax-adaptor/pull/51 as a PR so that we can exclude there are other dates lying around.

lsh-0:
> What's inside those fields there, some placeholder?
yup.

'published' doesn't need a placeholder, it's always in the xml.
'statusDate' and 'versionDate' are only added as placeholders: https://github.com/elifesciences/bot-lax-adaptor/blob/develop/src/validate.py#L203

PR #47, opened 2016-11-29
Adding both 200 for reports and 301 for protected pages

giorgiosironi:


giorgiosironi:
@lsh-0 your URLs will be loaded in all environments after deploy.

PR #52, opened 2016-12-06
PUBLISH events now fail if the article json merge is invalid.

lsh-0:
* tests fixtures updated. 
* fixtures are now copied from the patched/ dir rather than the previous 'dummyelife-foo.xml.json' files.


giorgiosironi:
Just checking the image URLs in end2end, then we can merge this.

lsh-0:
Please merge when you've done what you need to do

giorgiosironi:
Let's go.

PR #54, opened 2016-12-09
Bugfix update fragment

lsh-0:


giorgiosironi:
We will cover this bug at the end2end level, but is it possible to include a project-level test too?

lsh-0:
yes, it's why I haven't merged this PR yet.

lsh-0:
test added

PR #55, opened 2016-12-14
bugfix, article list now returns the correct title in the results.

lsh-0:
took the opportunity to split the `logic.latest_article_versions` function into two smaller functions. original function interface has been preserved as a simple wrapper

### DIFF
--- a/src/publisher/api_v2_views.py
+++ b/src/publisher/api_v2_views.py
@@ -62,9 +62,9 @@ def article_list(request):
     try:
         kwargs = request_args(request)
         kwargs['only_published'] = not authenticated
-        results = logic.latest_article_versions(**kwargs)
+        total, results = logic.latest_article_version_list(**kwargs)
         struct = {
-            'total': len(results), # TODO: probably wrong now.
+            'total': total,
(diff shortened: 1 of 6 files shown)


giorgiosironi:
Failed because:
```
08:58:36 [54.89.201.180] out:   File "/srv/lax/src/publisher/tests/test_ajson_ingest.py", line 455, in test_ingest_from_cli
08:58:36 [54.89.201.180] out:     self.assertEqual(result['datetime'][:20], expected_datetime[:20])
08:58:36 [54.89.201.180] out: AssertionError: u'2016-12-15T08:57:45Z' != '2016-12-15T08:57:46Z'
```

giorgiosironi:
Options:
- these should be compared by subtracting their timestamps, asserting that the difference is <= 1 or 2 seconds (a chosen tolerance)
- `@patch(datetime.datetime.now)` on top of the test to stop the flow of time during the test

lsh-0:
yeah - sorry. we've hit this one before, right?

both solutions are good. I'll open a jira ticket

lsh-0:
http://jira.elifesciences.org:8080/browse/ELPP-1945

PR #58, opened 2016-12-16
tweak to the fragment logic

lsh-0:
* introduced some better naming for the operations in `fragment_logic.py` 
* removed the boolean responses.
* added `set_article_json` that both sets and unsets article json depending on it's validity


lsh-0:
* roughed out the revalidation code
* added handling for articles with no fragments (StateError)
* moved the StateError class into utils. 
* gave the ArticleFragment class a better repr

need to turn the output of the `revalidate` function into a useful metric. attached it's output over a database of mixed content: [revalidate-results.txt](https://github.com/elifesciences/lax/files/657956/revalidate-results.txt)

PR #59, opened 2016-12-23
added complexity checker to test script

lsh-0:
this branch introduces a cyclometric complexity check. it scores functions, modules and overall and will fail the test if the function/module/overall scores exceed the thresholds.

I'm ignoring the `article_check` management script as it's something of a once-off and will be removed very soon.

I'm ignoring the `eif_ingestor.py` import script because EIF is no longer supported.

```
ERROR:xenon:block "src/publisher/ajson_ingestor.py:37 _ingest" has a rank of C
ERROR:xenon:block "src/publisher/ajson_ingestor.py:155 _publish" has a rank of C
ERROR:xenon:module 'src/publisher/ejp_ingestor.py' has a rank of B
ERROR:xenon:module 'src/publisher/ajson_ingestor.py' has a rank of B
```

This leaves a big chunk of complexity in the `ajson_ingestor.py` import script and something in the `ejp_ingestor.py` script. See? even with the best intentions these import scripts get complex very quickly!

There is code already for the `ajson_ingestor.py` to reduce the complexity here: https://github.com/elifesciences/lax/tree/feat-clearer-ingest-busrules


giorgiosironi:
Good idea, I thought Pylint was already doing some of these calculations.

lsh-0:
I guess those checks are more hygiene checks. Simpler to do, easy to implement, catch some code smells early on.

lsh-0:
might shift the cc check before the running of the tests. it's fast like pyflakes

PR #67, opened 2017-01-11
lax now stores the article-json returned from validation

lsh-0:
... (with formatted dates) rather than the one sent to validation

### DIFF
--- a/src/publisher/tests/test_fragment_logic.py
+++ b/src/publisher/tests/test_fragment_logic.py
@@ -77,6 +77,7 @@ class FragmentLogic(BaseCase):
 
 class FragmentMerge(BaseCase):
     def setUp(self):
+        # poa, published 2016-08-16T00:00:00Z
         self.ajson_fixture = join(self.fixture_dir, 'ajson', 'elife-16695-v1.xml.json')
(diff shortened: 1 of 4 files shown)


src/publisher/tests/test_fragment_logic.py
```diff
@@ -115,20 +117,66 @@ def test_valid_merge_updates_article_version_fields(self):
...
+        _, _, av2 = ajson_ingestor.ingest_publish(data)
+
+        # nothing should have changed
+        av2 = self.freshen(av)
+        self.assertEqual(expected, av2.article_json_v1['statusDate'])
+
+    def test_merge_sets_status_date_correctly_vor_v2(self):
+        "statusDate for a v2 vor is correctly set: earliest POA if POA, earliest VOR if VOR"
+        # load v2 vor
+        fixture = join(self.fixture_dir, 'ajson', 'elife-16695-v2.xml.json')
```
giorgiosironi: looks like some of the duplicated test code could be extracted in a `self.givenAPoaArticle()` or similar

lsh-0: I don't mind repeated code in test cases - I'd rather it were all painfully explicit and stupid, rather than clever or optimised.

PR #75, opened 2017-02-02
settings.py, db keeps it's connection open for 60seconds

lsh-0:


lsh-0:
lets see what happens :)

PR #78, opened 2017-02-08
Actually saving the new article JSON when adding a fragment through the API

giorgiosironi:
Updates all versions of an article.

### DIFF
--- a/src/publisher/api_v2_views.py
+++ b/src/publisher/api_v2_views.py
@@ -99,7 +99,7 @@ def article_version(request, id, version):
     try:
         # TODO: this seems wrong, should be 'not authenticated'
         # since you are already at it, test at the HTTP level also the other requests
-        av = logic.article_version(id, version, only_published=authenticated)
+        av = logic.article_version(id, version, only_published=not authenticated)
         return Response(logic.article_json(av), content_type=ctype(av.status))
     except models.ArticleVersion.DoesNotExist:
         raise Http404()
--- a/src/publisher/api_v2_views.py
+++ b/src/publisher/api_v2_views.py
@@ -51,6 +51,7 @@ def request_args(request, **overrides):
 #
 
 def is_authenticated(request):
+    # this header is never set, but only for this API because on /articles/42 it works
     val = request.META.get(settings.KONG_AUTH_HEADER)
     #LOG.info("authenticated? %s type %s" % (val, type(val)))
     return val or False
@@ -107,6 +108,7 @@ def article_version(request, id, version):
 
 @api_view(['POST'])
 def article_fragment(request, art_id, fragment_id):
+    #LOG.info('api_v2_views.article_fragment')
     only_published = is_authenticated(request)
     try:
         reserved_keys = [XML2JSON]
--- a/src/publisher/api_v2_views.py
+++ b/src/publisher/api_v2_views.py
@@ -108,7 +108,6 @@ def article_version(request, id, version):
 
 @api_view(['POST'])
 def article_fragment(request, art_id, fragment_id):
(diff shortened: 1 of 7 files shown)


src/publisher/api_v2_views.py
```diff
@@ -107,16 +110,18 @@ def article_version(request, id, version):
...
         reserved_keys = [XML2JSON]
         ensure(fragment_id not in reserved_keys, "that key is taken")
         with transaction.atomic():
-            av = logic.most_recent_article_version(art_id, only_published)
-            data = request.data
-            frag, created, updated = fragment_logic.add(av.article, fragment_id, data, update=True)
-            ensure(created or updated, "fragment was not created/updated")
-            fragment_logic.merge_if_valid(av, quiet=False)
+            avs = logic.article_version_list(art_id, only_published=False)
+            for av in avs:
```
lsh-0: This was not how fragments were intended to work

A fragment can be attached to an article or an articleversion. 
if a fragment exists at the article level, it gets applied to all article versions. 
if a fragment exists at the articleversion level, it only gets applied to that articleversion.

what you are now doing is adding the same fragment many times at the articleversion level.

giorgiosironi: So if a new version gets created it won't have that fragment, which is not what we want, I see the bug.
The original behavior was using the latest version, which is also not what we want. How do we attach the fragment to the article instead?

lsh-0: if you create a fragment, attach it to an article and don't specify a version, it gets merged for *all* article versions.

when the `xml->json` fragment is created, it's attached to a specific article version.

this way we can have article-global overrides and articleversion-specific overrides using fragments.

if it's working for now, that's fine, but it is on my radar to fix.

giorgiosironi: the current behavior is then that if a `vN` article version is published, the article "loses" the fragment because it's attached to `v1`, ..., `v(N-1)` only. I was wondering if there was a way to attach it to the article, but since it seems there is not a way to do that, we need a Jira ticket.

lsh-0: > but since it seems there is not a way to do that, we need a Jira ticket.

there is a way to do it.

the `version` attr in the fragment model can be `None`. if `None`, then it applies to the whole article, not a specific article version.

I'll create the ticket and fix this up.

lsh-0: http://jira.elifesciences.org:8080/browse/ELPP-2321

src/publisher/tests/test_api_v2_views.py
```diff
@@ -310,30 +316,58 @@ def setUp(self):
...
         self.c = Client()
+        self.ac = Client(**{
+            mware.CGROUPS: 'admin',
+        })
 
     def test_add_fragment(self):
         "a POST request can be sent that adds an article fragment"
         key = 'test-frag'
         url = reverse('v2:article-fragment', kwargs={'art_id': self.msid, 'fragment_id': key})
-        fragment = {'title': 'pants-party'}
```
lsh-0: you don't like the `pants-party` ? heh. an old boss used that as his 'foobar'. I have no idea what it means, but it's something that keeps on being funny to me

src/publisher/api_v2_views.py
lsh-0: and new PR: https://github.com/elifesciences/lax/pull/93

src/publisher/tests/test_api_v2_views.py
giorgiosironi: I think it's https://media0.giphy.com/media/VdU1mKwghuoak/200_s.gif

PR #79, opened 2017-02-09
Makes ordering deterministic

giorgiosironi:
Right now article latest versions are ordered by their published date. When more than one article has the same date, the order is not deterministic and paging through /articles results in duplicate items.

### DIFF
--- a/src/publisher/logic.py
+++ b/src/publisher/logic.py
@@ -145,7 +145,7 @@ def latest_unpublished_article_versions(page=1, per_page=-1, order='DESC'):
     start = (page - 1) * per_page
     end = start + per_page
     order_by = ['datetime_published', 'article__manuscript_id']
-    if order is 'DESC':
+    if order == 'DESC':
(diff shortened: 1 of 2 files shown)


giorgiosironi:
I also found out `order=desc` doesn't work for authenticated users, that's the additional commit.

lsh-0:
argh. well done. I thought the problem lay in the nested join and was tearing my beard out trying to figure just wtf I'd done wrong. order + paging didn't occur to me.

giorgiosironi:
To be fair `is` used on a string here https://github.com/elifesciences/lax/pull/79/commits/6188837b8b3e69defb981514165c75ebefd78264#diff-9592b6d80ed9da65b850b9605befa384L148 should have been the job of Pylint/Pyflakes to find

lsh-0:
eh, it's valid and I don't expect the linters to understand the intent. it may come up in some of the soft warnings emitted by pylint that we ignore

PR #90, opened 2017-02-20
Feat related articles

lsh-0:


src/core/settings.py
```diff
@@ -241,8 +243,9 @@ def _load_api_raml(path):
 
 INGESTION_LOG_NAME = 'ingestion-%s.log' % PROJECT_NAME
 
-LOG_FILE = join('/var/log/', LOG_NAME) # ll: /var/log/lax.log
-INGESTION_LOG_FILE = join('/var/log/', INGESTION_LOG_NAME) # ll: /var/log/lax.log
+LOG_DIR = '/var/log/' if ENV != DEV else './'
```
giorgiosironi: there was some problem here with the smoke tests being broken on the Vagrant VM, let me retrieve what was that

giorgiosironi: https://github.com/elifesciences/lax/commit/edf8bde3c9a9674728b2445c1b52203fa5414f4f essentially the uwsgi-lax process couldn't write file like `/srx/lax/lax.log` in the Vagrant VM; how do we make it work both in the VM and in local checkouts of Lax?

lsh-0: ok, I'll investigate why uwsgi can't write to `/srv/lax/lax.log`.

lsh-0:
this is ready to merge now. 

@giorgiosironi , @thewilkybarkid , an article referencing another elife article as related content will fail if that other article doesn't exist. now, lax--prod has the ejp-adaptor feeding it article stubs from EJP, so that failure is unlikely to ever happen there, but in the end2end environment where these stubs don't exist, a dummy article related to another article that doesn't exist will cause ingestion to fail (unless forced).

I *could* create the Article stub automatically if it doesn't exist. Suggestions?

giorgiosironi:
Thank for thinking about testability. :cookie: 

First, what does it mean to fail? Is it an error when loading the `/articles/{id}/related` API or an error with a larger impact like failing the ingestion? 

Second, I can manipulate the input XML when generating the articles, but it's difficult to guess some existing article ids. Therefore creating the stub automatically could be a good idea and also covers the unlikely possibility that production encounters this situation. I would like to see a sample `/related` response resulting from these stubs being created automatically, just to check what we are going to encounter.

lsh-0:
point 1 - a fail means the article doesn't get ingested unless the ingest operation is forced, failing with a StateError.

point 2 - creating a stub in the absence of an article means the ingestion will pass but the `/related` endpoint, because the stub doesn't have an ArticleVersion, no publication status and certainly no article-json, will not return any mention of that relationship.

I slept on it and I think a stub is definitely the right way to go. We're being fed from EJP at the moment but in another system that may not be the case. Ingest with actually bad relationships will have the bad relationship cleaned up but the bad stub will still exist. I'll defer handling their cleanup for the time being but will give them some identifier so we can find them in the future

lsh-0:
> Thank for thinking about testability.

oh, and of course!

lsh-0:
I've created an option in settings.py to allow stubs to be created. It's `false` by default. if we need to turn it on, we just alter the `app.cfg` config file.

src/core/settings.py
lsh-0: I can replicate this by removing `/var/log/lax.log` and restarting the `uwsgi-lax` service. I'll see if upstart can create the log file with the right permissions before dropping perms

lsh-0: ok, I can fix this by modifying the upstart script. going to miss upstart when it's gone :(

lsh-0: handled here: https://github.com/elifesciences/lax-formula/commit/8e8b5d637ef5b4e071fe20e994b18885b8a689ac

it's a neat little trick, we might want to consider it for some of our other projects

PR #93, opened 2017-02-21
tweak to the fragment POST logic.

lsh-0:


src/publisher/api_v2_views.py
```diff
@@ -98,8 +98,7 @@ def article_version(request, id, version):
     "returns the article-json for a specific version of the given article ID"
     authenticated = is_authenticated(request)
     try:
-        # TODO: this seems wrong, should be 'not authenticated'
-        # since you are already at it, test at the HTTP level also the other requests
+        # TODO: test at the HTTP level also the other requests
```
giorgiosironi: comment was outdated, right

src/publisher/api_v2_views.py
```diff
@@ -129,12 +128,16 @@ def article_fragment(request, art_id, fragment_id):
         reserved_keys = [XML2JSON]
         ensure(fragment_id not in reserved_keys, "that key is taken")
         with transaction.atomic():
+            article = models.Article.objects.get(manuscript_id=art_id)
+            data = request.data
+            frag, created, updated = fragment_logic.add(article, fragment_id, data, update=True)
```
giorgiosironi: right so we only need to do this once as all `av` were pointing to the same article

giorgiosironi:
Seems a good simplification. Does it have an impact on behavior that we can pin down with a new test (initially failing but green after the change)?

lsh-0:
I'm not sure what the original problem was that you were seeing that caused you to touch that code, so I can't test for it.

giorgiosironi:
If anything emerges from end2end I'll port back a smaller test case, no problem.

PR #95, opened 2017-02-24
Updated bot-lax-adaptor to: Merge pull request #124 from elifesciences/feature-elifetools-update

elife-alfred-user:
I have run http://ci--alfred.elifesciences.org/job/dependencies-lax-update-bot-lax-adaptor/13/ which resulted in this pull request.


giorgiosironi:
Merged the previous one so this conflicts... will wait for the next update.

thewilkybarkid:
This change requires a rebuild of all articles (changes the most common reference type). Can we get it in manually and I'll update Journal to support the change at the same time.

giorgiosironi:
Ok, I'll open a new PR.

giorgiosironi:
The problem is we can only rebuild for published articles, we have no way of doing so for unpublished articles at the moment.

giorgiosironi:
A way that comes to mind is for Lax to store the XML URL of each article version, so that it can be asked about it by the bot-lax-adaptor when doing the rebuild (rather than only using the XML of published articles from elife-article-xml).

PR #99, opened 2017-02-28
added a status report 

lsh-0:
it will output a helpful list of articles that are unpublished/invalid/unpublished+invalid

bot-lax will call this command to get the list of unpublished articles in order to backfill


src/reports/logic.py
```diff
@@ -12,6 +13,68 @@ def take(n, iterable):
...
+
+    def get_loc(av):
+        try:
+            return av.article.articlefragment_set.get(type=models.XML2JSON).fragment['article']['-meta']['location']
+        except models.ArticleFragment.DoesNotExist:
+            return 'no-article-fragment'
+        except KeyError:
+            return 'no-location-stored'
+
+    def avi(av):
```
giorgiosironi: some docstrings to understand what `avi` and `avil` is the acronym of would be helpful

giorgiosironi:
https://github.com/elifesciences/bot-lax-adaptor/pull/126 depends on this, so better get started with merging.

PR #109, opened 2017-03-06
ingest will now fail an invalid article. 

lsh-0:
* updated fixtures
* removed all those fixtures prefixed with 'dummy' as we now have 100% valid article scrapes (with minor, necessary, patching). 
* added my script to update fixtures
* updated fixtures means we now have relations being ingested and stubs created. added a shim to avoid this and cut down on some code. open to better solutions.


lsh-0:
er - it's passing, but Jenkins is dead with:
```
GitHub has been notified of this commit’s build result

java.lang.IllegalStateException: Can not access the context of a running build
```

giorgiosironi:
Not debugged the problem in depth so I am rerunning to see if it reproduces.

lsh-0:
gave it another kick, seems to work now.

ticket is here: https://elifesciences.atlassian.net/browse/ELPP-2386

src/publisher/tests/fixtures/ajson/gen-fixtures.sh
```diff
@@ -0,0 +1,12 @@
...
+set -xuv
+home=$(pwd)
+cd /home/luke/dev/python/bot-lax-adaptor
+for fname in elife-16695- elife-01968- elife-20125- elife-20105- elife-12215-v1
+do
+   for path in ./article-xml/articles/$fname*
+   do
+     ./scrape-article.sh "$path" > $home/${path##*/}.json
+   done
+done
```
giorgiosironi: nice automation

src/publisher/ajson_ingestor.py
```diff
@@ -78,7 +78,7 @@ def _ingest(data, force=False):
         # only update the fragment if this article version has *not* been published *or* if force=True
         update_fragment = not av.published() or force
         fragments.add(av, XML2JSON, data['article'], pos=0, update=update_fragment)
-        fragments.set_article_json(av, quiet=True)
+        fragments.set_article_json(av, quiet=force) # article-json validation occurs here
```
giorgiosironi: so if `force` is specified the article would still be ingested? Silent corrections do that, but we want to validate the JSON on ingest for them too, to catch the problem as soon as possible

src/publisher/tests/fixtures/ajson/gen-fixtures.sh
lsh-0: these `${path##*/}` crazy things scare me, but it seems to work.

src/publisher/ajson_ingestor.py
lsh-0: > but we want to validate the JSON on ingest for them too

warnings will still be emitted to logs that are sent to loggly and then emailed to us (current situation).

giorgiosironi: what's the end result there? A silent correction where the new JSON is ignored or does it actually break the article making it invalid? Can we do better than both cases and leave the old version in place returning an error?
Or is it because `force` should save the result no matter what? In that case silent corrections should use a different parameter as their intent is to update the article in safety, not no matter what's the result.

lsh-0: > Or is it because force should save the result no matter what?
correct. if you force something, you live with the results.

silent corrections are us saying: "modify a published article and don't tell anyone!", which feels pretty wrong. `force` neatly bypasses the rules, but there *is* that window for invalid ajson to get in.

to do things correctly here would require overhauling eLife's entire production process (no such thing as a 'silent correction', ever), and removing the `force` semantics seems very impractical (forcing an action may still fail), but a compromise could be:

`quiet=(settings.VALIDATE_ON_FORCE or force)`

so, if `VALIDATE_ON_FORCE` is `False`, and `force` is `True`, we preserve the very useful semantics of forcing an action.

if `VALIDATE_ON_FORCE` is `True`, a forced action may still fail BUT we have the certainty invalid articles can't sneak in this way, leaving us with just a changing spec causing invalidity.

I'm just trying to think if this flexibility is even useful to us or others ...

giorgiosironi:
I don't know about the flexibility, but it feels good to know that a silent correction cannot break the article page (that's what the schema is for). Place some safety on that process.

lsh-0:
cool. so, as I mentioned in the jira ticket, I want to leave it to you guys to merge this one as it will affect production and they should have somebody in the office who knows what is going on and can debug it during your day.

giorgiosironi:
Let's go.

PR #111, opened 2017-03-07
Do not install on article ingests

giorgiosironi:
Today article ingests got stuck in end2end (possibly also prod) when pulling stuff down from the Internet:
```
  808 ?        Ssl    0:03 python src/adaptor.py --type sqs
 1795 ?        S      0:00  \_ /bin/bash /srv/lax/manage.sh ingest --ingest --id 2258964234223315893 --version 1
 1813 ?        S      0:00      \_ /srv/lax/venv/bin/python3.5 /srv/lax/venv/bin/pip install -r requirements.txt
 1820 ?        S      0:00          \_ /usr/bin/python /usr/bin/hg clone --noupdate -q https://bitbucket.org/lskibinski/rfc3339 /tmp/pip-6cj82ib7-build
```

### DIFF
--- a/manage.sh
+++ b/manage.sh
@@ -2,5 +2,9 @@
 # @description convenience wrapper around Django's runserver command
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 cd $DIR
-source install.sh > /dev/null
+
+# this starts to reinstall everything in production everytime an article is ingested:
+#source install.sh > /dev/null
+
+source venv/bin/activate
 ./src/manage.py $@


giorgiosironi:
Incidentally this should also make the ingests and end2end tests faster

giorgiosironi:
The build, however, takes forever now that they are slow, that's expected.

lsh-0:
hey man - this is a big change. we rely on `manage.sh` ensuring the venv is correct and migrations applied. if you have a scenario where you can guarantee the venv exists and doesn't need to be installed then a better solution would have been passing in a flag to skip sourcing `install.sh`

giorgiosironi:
Was in an emergency as article ingests stopped working for a couple of hours; I saw the new flag and opened pull requests to use it, good to have the option.

lsh-0:
yeah, I get that happens. we need to be vigilant against these creeping exceptions-to-the-rule.

PR #126, opened 2017-03-15
removed non-deterministic db ordering issue when merging fragments

lsh-0:


lsh-0:
```
lax=# \d publisher_articlefragment
                                          Table "public.publisher_articlefragment"
         Column          |           Type           |                               Modifiers                                
-------------------------+--------------------------+------------------------------------------------------------------------
 id                      | integer                  | not null default nextval('publisher_articlefragment_id_seq'::regclass)
 fragment                | text                     | not null
 article_id              | integer                  | not null
 position                | smallint                 | not null
 type                    | character varying(25)    | not null
 datetime_record_created | timestamp with time zone | not null
 version                 | smallint                 | 
```

https://www.postgresql.org/docs/9.1/static/datatype-datetime.html:

|Name|Storage Size|Description|Low Value|High Value|Resolution|
|--- |--- |--- |--- |--- |--- |
|timestamp [ (p) ] with time zone|8 bytes|both date and time, with time zone|4713 BC|294276 AD|1 microsecond / 14 digits|

PR #133, opened 2017-03-20
Feat ingest history dates

lsh-0:
way back in the beginning of lax I made the mistake of dumping all of the EJP stuff into the Article table. all the revision and decision information belong in a separate table.

more recently I attempted to set up database logging to better capture what operations had been performed on an article, but it never made the light of day and probably wasn't the best solution anyway. it wouldn't have been any less noisy than Loggly in trying to pin down what had happened to an article.

this PR introduces that event table missing since the beginning and adds the first set of events to it: received, accepted, ingest action, publish action

the received and accepted dates are pulled from the bot-lax adaptor if present.

two caveats:
* silent corrections on receive/accept dates will have multiple values in there. if the updated values are backdated, the logic will only ever pick up the most recent one
* if an error happens and the transaction is rolled back, I won't have the event entry to say an ingest attempt happened, but failed (this can probably be overcome)


lsh-0:
hrm - it's failing because `--ci` is using an older database where the `0041` migration *was* run.

giorgiosironi:
What we do on other projects that have a database is to create a fresh copy during any `ci` build (but not on `end2end`, or `prod` of course). Since pull requests are build on `ci`, you may find the existing database in strange states especially if they fail.

This can be done by tweaking the formula or by adding a command to `project_tests.sh`, let me know what you prefer.

lsh-0:
> This can be done by tweaking the formula or by adding a command to project_tests.sh, 
> let me know what you prefer.

it's not a problem. what I did, squashing the migration history, then tweaking it, then re-creating a migration and pointing it to the squashed migration, was such a once-off and relatively complex/magical that I'm surprised it didn't fail harder than it did.

src/publisher/ajson_ingestor.py
```diff
@@ -31,51 +31,82 @@
...
+
+    except KeyError as err:
+        raise ValueError("failed to scrape article data, couldn't find key %s" % err)
+
+
+#
+#
+#
+
+def _ingest(data, force=False) -> models.ArticleVersion:
```
lsh-0: this is interesting. a few weeks ago I came across typing in python 3 that had just never crossed my radar before. I'm looking for a static analysis tool that I can run as part of the linting and possibly replace pylint. `mypy` is a strong contender: http://mypy-lang.org/

PR #147, opened 2017-03-27
Added /ping to /api/v2

giorgiosironi:
To be targeted by journal in its status page and possibly by monitoring.

### DIFF
--- a/src/publisher/api_v2_urls.py
+++ b/src/publisher/api_v2_urls.py
@@ -2,6 +2,7 @@ from django.conf.urls import url
 from . import api_v2_views as views
 
 urlpatterns = [
+    url(r'/ping$', views.ping, name='ping'),
     url(r'articles$', views.article_list, name='article-list'),
     url(r'articles/(?P<id>\d+)$', views.article, name='article'),
     url(r'articles/(?P<id>\d+)/versions$', views.article_version_list, name='article-version-list'),
--- a/src/publisher/api_v2_urls.py
+++ b/src/publisher/api_v2_urls.py
@@ -2,7 +2,7 @@ from django.conf.urls import url
 from . import api_v2_views as views
 
 urlpatterns = [
-    url(r'/ping$', views.ping, name='ping'),
+    url(r'^ping$', views.ping, name='ping'),
     url(r'articles$', views.article_list, name='article-list'),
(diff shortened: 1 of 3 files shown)


src/publisher/api_v2_urls.py
```diff
@@ -2,6 +2,7 @@
 from . import api_v2_views as views
 
 urlpatterns = [
+    url(r'/ping$', views.ping, name='ping'),
```
lsh-0: no leading slash

src/publisher/api_v2_views.py
```diff
@@ -57,6 +58,14 @@ def is_authenticated(request):
     return val or False
 
 @api_view(['GET'])
+@renderer_classes((StaticHTMLRenderer,))
+def ping(request):
+    "returns a test response"
+    # produce standard authenticated header
+    is_authenticated(request)
```
lsh-0: this isn't doing anything. `is_authenticated` has no side effects

src/publisher/api_v2_views.py
```diff
@@ -57,6 +58,14 @@ def is_authenticated(request):
...
+    return Response('pong', content_type='text/plain; charset=UTF-8', headers={'Cache-Control': 'must-revalidate, no-cache, no-store, private'})
```
lsh-0: if these cache headers are different for the /ping endpoint than for all the others, then this is fine, otherwise, caching headers are handled in custom middleware in `core.middleware` (beneath the kong authentication middleware) and this can be removed.

src/publisher/tests/test_api_v2_views.py
```diff
@@ -116,6 +116,13 @@ def setUp(self):
             mware.CGROUPS: 'admin',
         })
 
+    def test_ping(self):
+        resp = self.c.get(reverse('v2:ping'))
+        self.assertEqual(resp.status_code, 200)
+        self.assertEqual(resp.content_type, 'text/plain; charset=UTF-8')
+        self.assertEqual(resp['Cache-Control'], 'must-revalidate, no-cache, no-store, private')
```
lsh-0: again, double handling if /ping cache headers are no different from regular caching headers.

giorgiosironi:
Changes made

giorgiosironi:
For some reason this failed when in `develop`, trying to reproduce locally.

PR #149, opened 2017-04-03
fixed missing leading arrows in urls.

lsh-0:
this made certains urls possible when they shouldn't have been.


giorgiosironi:
Cheers, :+1: this first happened to me when I was trying to load `/articles/42/related-articles` and getting the whole article list as a result because the correct URL was `/articles/42/related`

lsh-0:
yeah, I could have *sworn* it was simply concatenating the regex behind the scenes, but I guess I was wrong. I checked all other instances and I've been inconsistent in it's application, so I tidied everything up

PR #164, opened 2017-04-18
Feature: trigger article update events from admin

lsh-0:
this PR:
* sends article update events to aws if articles or articleversions are changed or deleted via Django's admin interface
* sends article update events to aws if fragments are added or deleted via the fragment API
* re-generates the article-json after a change to the article or it's versions via the admin interface

### DIFF
--- a/src/publisher/admin.py
+++ b/src/publisher/admin.py
@@ -29,7 +29,6 @@ class ArticleAdmin(admin.ModelAdmin):
         super(ArticleAdmin, self).delete_model(request, art)
         aws_events.notify(art)
 
-    # this may cause multiple events to be sent
     def save_related(self, request, form, formsets, change):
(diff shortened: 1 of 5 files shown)


src/publisher/admin.py
```diff
@@ -1,5 +1,5 @@
 from django.contrib import admin
-from . import models
+from . import models, aws_events
```
giorgiosironi: well, you shouldn't strictly need to name your module after the specific technology `aws`, `events` would make a nice pair with `models` (which is not `postgresql_models`)

lsh-0: I think there is (or was?) a module called events already for something else. It was once called sns_events or something similar. when we get different types of events that need to be sent out, I'll deal with the  refactor then. It's pretty benign right now ;)

PR #165, opened 2017-04-19
added more fragments to the article snippet

lsh-0:
getting some annoying errors locally I can't pin down. opening PR to see if they can be replicated remotely.

update: can't replicate it with sqlite. I think it's just  psql + green's multithreading

### DIFF
--- a/src/publisher/fragment_logic.py
+++ b/src/publisher/fragment_logic.py
@@ -92,10 +92,14 @@ def extract_snippet(merged_result):
         return None
     # TODO: derive these from the schema automatically somehow please
     snippet_keys = [
+        # https://github.com/elifesciences/api-raml/blob/develop/src/snippets/article-vor.v1.yaml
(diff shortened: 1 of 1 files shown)


giorgiosironi:
I think that originally only a subset of `image` was appearing in the snippet (some subfields) but this has been simplified now and there should be only one image for an article.

lsh-0:
um - I think I can avoid that. if I can't, I'll do a backfill today

lsh-0:
yeah, a `./manage.sh revalidate` should do the trick.

lsh-0:
> I think that originally only a subset of image was appearing in the snippet (some subfields) 
> but this has been simplified now and there should be only one image for an article.

I actually encountered this trying to figure out how I managed to make an article invalid in the admin the other day -  the `image` field has become more complex. before it was just a single `uri` type field, now it's a sub-struct with `uri` within it and a lot of other fields. I was using an old database dump and the `image` field with a single value now failed to validate.

PR #200, opened 2017-05-24
Sending events on related articles

giorgiosironi:
Relations are explicitly passed in, while in these tests they are normally removed. We add events for the related articles to be sent together with the main one being ingested/published, even if the related articles are stubs.

Two new fixtures are introduce, related Homo Naledi articles.

### DIFF
--- a/src/publisher/aws_events.py
+++ b/src/publisher/aws_events.py
@@ -117,7 +117,7 @@ def notify(art, **overrides):
     try:
         msg = {"type": "article", "id": art.manuscript_id}
         msg_json = json.dumps(msg)
-        LOG.debug("writing message to event bus", extra={'bus-message': msg_json})
+        LOG.info("writing message to event bus", extra={'bus-message': msg_json})
         event_bus_conn(**overrides).publish(Message=msg_json)
     except ValueError as err:
         # probably serializing value
--- a/src/publisher/aws_events.py
+++ b/src/publisher/aws_events.py
@@ -1,6 +1,8 @@
 import json
 from django.conf import settings
 import boto3
+from publisher import relation_logic as relationships
+from publisher.utils import lmap
 
 import logging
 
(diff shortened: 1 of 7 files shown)


lsh-0:
hey man, I refactored your work a bit, I think the transaction handling within the unittests threw you a bit and I dealt with it in the metrics project a while back.

feel free to merge if you're happy I've captured your intent.

src/publisher/aws_events.py
```diff
@@ -16,112 +18,35 @@ def sns_topic_arn(**overrides):
...
-        LOG.info("deleting queue")
-        if queue:
-            queue.delete()
-
-        if subscription:
-            subscription.delete()
-
-        if perm_label:
-            topic.remove_permission(Label=perm_label)
-'''
```
giorgiosironi: I like deleted code :+1:

src/publisher/aws_events.py
```diff
@@ -16,112 +18,35 @@ def sns_topic_arn(**overrides):
...
+    "notify event bus when this article or one of it's versions has been changed in some way"
     if settings.DEBUG:
-        LOG.debug("application is in DEBUG mode, not notifying anyone")
+        LOG.debug("application is in DEBUG mode, no notifications will be sent")
         return
     try:
         msg = {"type": "article", "id": art.manuscript_id}
         msg_json = json.dumps(msg)
-        LOG.info("writing message to event bus", extra={'bus-message': msg_json})
+        LOG.debug("writing message to event bus", extra={'bus-message': msg_json})
```
giorgiosironi: I still think that we should have a log of all events being sent, for practical debugging; I am not sure `debug` is captured in production.

giorgiosironi:
Same behavior, implementation more in tune with the rest of Lax, thanks.

src/publisher/aws_events.py
lsh-0: it was good code, but we have the infrastructure now to properly test and debug this without having to create adhoc topics and queues

PR #213, opened 2017-06-02
Feature: new backfill

lsh-0:
adding support for the new backfill happening in bot-lax.

* disabled commands: `repop`, `output`,  `fix_updated_dates` - these were all temporary scripts and are now no longer used (and probably very dangerous)
* disabled command: `sns`, it sends article update notfications to sns from the command line. the amount of code is trivial and easily replicated in a repl if (ever) necessary.
* removed old 'published.csv' report that lives in the `observer` project
* removed two reports that gave some stats about articles published in the specified year broken down by article type - these are now trivial to create inside observer's metabase installation
* renamed `status` command in report app to `report`. the backfill script will be hitting this command to extract an index of articles from lax


src/publisher/fragment_logic.py
```diff
@@ -214,6 +214,21 @@ def delete_fragment_update_article(art, key):
         return set_all_article_json(art, quiet=False)
 
 #
+#
+#
+
+def location(av):
```
giorgiosironi: not sure yet where this is used, but doesn't it seem dangerous to return strings as error values while the valid value is a string too? What about using a tuple or custom exceptions?

src/reports/management/commands/report.py
```diff
@@ -0,0 +1,54 @@
...
+
+LOG = logging.getLogger(__name__)
+
+def avl2csv():
+
+    rs = models.ArticleVersion.objects \
+        .select_related('article') \
+        .defer('article_json_v1', 'article_json_v1_snippet') \
+        .order_by('article__manuscript_id', 'version') \
+        .all()
```
giorgiosironi: I think this may find stubs like articles coming from EJP but not yet ingested, or articles that have been related to an existing one but neither yet ingested. This is not necessarily a problem as long as we are aware of these cases, which won't be backfilled.

src/publisher/fragment_logic.py
lsh-0: I don't see danger but I do understand where you are coming from. This isn't new code, it's been shifted from the status report that I removed and into fragment logic. There aren't many cases where we don't have the location any more, but dealing with the string 'no-location' in bash is very easy. If something else uses it in future and needs it to be noisy, we can raise exceptions then. Doing a curl on the string 'no-location' produces a wonderfully clear error btw ;)

src/reports/management/commands/report.py
lsh-0: > Or maybe they have no articleversion so they won't come out.

correct. an ArticleVersion object is created upon first ingest from the bot.

src/publisher/fragment_logic.py
giorgiosironi: Yeah I see the bash usage later, in bash everything is a string :)

lsh-0:
finished a test of the new backfill. I'm satisfied. this should be merged in at the same time as the bot-lax branch to avoid not being able to backfill. this PR removes functionality the current backfill relies on.

giorgiosironi:
Needs to be updated with the bot-lax-adaptor master before merging.

PR #228, opened 2017-06-23
Bugfix: ELPP-2856

lsh-0:
this is a small part of the set of changes needed to make the backfill a little more robust

### DIFF
--- a/src/reports/tests/test_reports.py
+++ b/src/reports/tests/test_reports.py
@@ -1,7 +1,8 @@
+from unittest.mock import patch
 import re
 from os.path import join
 from publisher.tests import base
-from publisher import eif_ingestor, logic as publogic, utils, ajson_ingestor, models
+from publisher import eif_ingestor, logic as publogic, utils, ajson_ingestor
 from reports import logic
 
 from django.test import Client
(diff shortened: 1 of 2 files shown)


lsh-0:
@giorgiosironi , there are bad `location` values in lax preventing the backfill from completing. I've added support for a case I don't think we can get around easily and a better error message for the cases where we have these bad values.

in this PR there is a (temporary) command called `fix-broken-locs` callable with `./manage.sh report fix-broken-locs` that will go through the article json and fix them up. I'll remove it eventually.

It has the `defer` clause in the object selection to prevent the article-json from being loaded until accessed. it causes more sql queries but won't bring the machine to it's knees like I did earlier.

src/reports/tests/test_reports.py
```diff
@@ -1,12 +1,82 @@
...
 from os.path import join
 from publisher.tests import base
-from publisher import eif_ingestor, logic as publogic, utils
+from publisher import eif_ingestor, logic as publogic, utils, ajson_ingestor
 from reports import logic
 
 from django.test import Client
 from django.core.urlresolvers import reverse
 
+class CMD1(base.BaseCase):
```
giorgiosironi: any better name than `CMD1`, `test_cmd1`...?

PR #244, opened 2017-07-14
Feature: eif removal

lsh-0:
this is an effort to remove some of the legacy code lax has accrued in it's short life.

* it removes v1 of the api
* it removes all support for 'eif' - the article json format that existed before article-json
* it replaces the article stub creator for testing with something that goes through the business logic rather than steps around it.


src/publisher/tests/base.py
```diff
@@ -39,11 +41,39 @@ def call_command(self, *args, **kwargs):
         stdout = StringIO()
         try:
             kwargs['stdout'] = stdout
-            call_command(*args, **kwargs)
+            dj_call_command(*args, **kwargs)
```
giorgiosironi: :dark_sunglasses:

src/publisher/tests/base.py
```diff
@@ -39,11 +41,39 @@ def call_command(self, *args, **kwargs):
...
+
+            'title': '[default]',
+            'version': 1,
+            'status': models.VOR,
+            'published': '2012-01-01T00:00:00Z'
+        }
+        struct.update(adata)        
+        delall(struct, ['journal']) # can't be serialized, not utilised anyway
+        
+        with self.settings(VALIDATE_FAILS_FORCE=False):
```
lsh-0: finally found a use for this settings. vindicated!

lsh-0:
I'm going to do a quick second pass on Monday and then merge this in then.

I used this tool: https://pypi.python.org/pypi/vulture  to help me hunt down some unused functions. unfortunately it's a bit noisy because of Django magic.

PR #251, opened 2017-07-21
Feat record all events

lsh-0:
next step in shifting some of the values captured from EJP out of the Article table into the ArticleEvent table.


src/publisher/models.py
```diff
@@ -266,44 +266,15 @@ class Meta:
         # lowest positions first (ASC), earliest creation date first (ASC)
         ordering = ('position', 'datetime_record_created')
 
+DATE_EJP_QC, DATE_EJP_DECISION = 'date-ejp-qc', 'date-ejp-decision'
```
giorgiosironi: can we give names to these fields/events that abstract EJP away from the model? Lax could be usable with any peer review system, not just with EJP

lsh-0: I agree with you ... but 'EJP' is pretty pervasive throughout the lax codebase. changing it here and not elsewhere seems hacky. I suggest a separate feature branch to replace any specific vendor naming.

to avoid a future data migration, I'll change the values added to the db, but not the symbols.

giorgiosironi: perfectly reasonable

PR #257, opened 2017-07-31
Feature: buffer aws events

lsh-0:
* an attempt to control the quantity of notifications sent by lax during a backfill.
* introduces ability to disable relationship creation
* slightly more useful logging messages in the `ingest` management command
* serial bulk ingest now available


src/publisher/aws_events.py
```diff
@@ -1,52 +1,108 @@
...
+    return sns.Topic(sns_topic_arn())
+
+#
+#
+#
+
+SAFEWORD = START = STOP = 'cacao'
+
+def defer(fn):
+    """calls function normally until safeword is received then buffers all requests until the safeword is called again.
```
giorgiosironi: This looks like it needs three primitives to call from outside:

- open()
- do()
- close()

rather than a single function; the next person passing from here will thank you

src/publisher/aws_events.py
```diff
@@ -1,52 +1,108 @@
+from ordered_set import OrderedSet
 import json
 from django.conf import settings
 import boto3
 from publisher import relation_logic as relationships
-from publisher.utils import lmap
+from functools import wraps
+
+from multiprocessing import SimpleQueue
```
giorgiosironi: `SimpleQueue` is for multiple threads writing to the same data structure I think. Why not just `Queue.Queue`?

src/publisher/aws_events.py
```diff
@@ -1,52 +1,108 @@
...
+    return wrapper
+
 
 #
 #
 #
 
-def notify(art, **overrides):
+@defer
+def notify(msid):
```
giorgiosironi: :+1: simpler to call

giorgiosironi: is it a string or an integer? zeropadded or not?

src/publisher/management/commands/ingest.py
```diff
@@ -164,7 +163,12 @@ def handle_many(print_queue, action, path, force, dry_run):
         LOG.info("found no article json at %r" % os.path.abspath(path))
         clean_up(print_queue)
         return
-    Parallel(n_jobs=-1)(delayed(job)(print_queue, action, path, force, dry_run) for path in ajson_file_list) # pylint: disable=unexpected-keyword-arg
+    try:
+        aws_events.notify(STOP) # defer sending notifications
+        Parallel(n_jobs=-1)(delayed(job)(print_queue, action, path, force, dry_run) for path in ajson_file_list) # pylint: disable=unexpected-keyword-arg
```
giorgiosironi: ah, that's why the concurrent queue; is the multiple job version used only during backfill or are there other scenarios?

src/publisher/aws_events.py
lsh-0: the manuscript ID is an integer. some systems pad it with zeroes, lax doesn't, except when sending it out into the world. if it were padded, I'd have `padded_msid` or similar. I think storing the padded msid is a bad idea and tried to get it changed at the beginning of the 2.0 overhaul but to no avail.

src/publisher/management/commands/ingest.py
lsh-0: multiple job version is used when ingesting a directory rather than a single file. the backfill calls lax with a directory of freshly scraped and validated article-json. being able to ingest a directory was very useful during development, it also cuts down on the overhead of calling the manage.sh script thousands of times

src/publisher/aws_events.py
lsh-0: lax ingests using Python's multiprocessing

src/publisher/aws_events.py
lsh-0: the `**overrides` was a leftover from a management function that wasn't being used. it allowed lax to listen to events from different environments.

src/publisher/aws_events.py
lsh-0: I don't think so. the mechanism is very simple, the aliases are there to make the intent even clearer. if you can implement this even more simply using a three function approach I'll include it.

lsh-0:
I'll write a test to do a dir ingest today and ensure it works as expected while multithreaded, but I'm not expecting any further changes.

src/publisher/management/commands/ingest.py
```diff
@@ -148,10 +149,8 @@ def handle_single(print_queue, action, infile, msid, version, force, dry_run):
 
 
 def job(print_queue, action, path, force, dry_run):
-    _, padded_msid, suffix = os.path.basename(path).split('-')
-    msid = int(padded_msid)
-    version = int(suffix[1])
     try:
+        msid, version = utils.version_from_path(path)
```
lsh-0: I've found a bug here. msid is returned as a string. fixing.

src/publisher/aws_events.py
giorgiosironi: This is what I meant: https://github.com/elifesciences/lax/pull/260

lsh-0: Thanks for that.

PR #265, opened 2017-08-11
uses native python3 virtualenv to create virtualenv

lsh-0:
all python3 apps should use the native virtualenv to create a virtualenv.

`virtualenv` will still be available for any python2 apps in 16.04 but should be considered deprecated. python2 is approaching EOL too.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -4,30 +4,26 @@ echo "[-] install.sh"
 
 . download-api-raml.sh
 
-# use the latest version of python3 we can find. 
-# on Ubuntu14.04 the stable version is 3.3, the max we can install is 3.6
-
-# ll: /usr/bin/python3.6
-#maxpy=$(which /usr/bin/python3* | grep -E '[0-9]$' | sort -r | head -n 1)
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -4,30 +4,26 @@ echo "[-] install.sh"
...
-
-# ll: /usr/bin/python3.6
-#maxpy=$(which /usr/bin/python3* | grep -E '[0-9]$' | sort -r | head -n 1)
-maxpy=/usr/bin/python3.5
-
-# ll: python3.6
-# http://stackoverflow.com/questions/2664740/extract-file-basename-without-path-and-extension-in-bash
-py=${maxpy##*/} # magic
+python=/usr/bin/python3.5
+py=${python##*/} # ll: python3.5
```
giorgiosironi: what does this expansion do? need a comment?

lsh-0: it's an bashism. you see the before, you see the after - I don't think it needs an explanation. It's actually more understandable now I've shifted `python` out of the fs lookup

PR #280, opened 2017-09-13
Feature: api v2 support

lsh-0:


src/publisher/negotiation.py
```diff
@@ -2,28 +2,40 @@
...
-    ('ArticleList', 'application/vnd.elife.article-list+json', [1]),
-    ('POAArticle', 'application/vnd.elife.article-poa+json', [1]),
-    ('VORArticle', 'application/vnd.elife.article-vor+json', [1]),
-    ('ArticleHistory', 'application/vnd.elife.article-history+json', [1]),
-    ('ArticleRelated', 'application/vnd.elife.article-related+json', [1]),
+    ('ArticleList', 'application/vnd.elife.article-list+json', [1, 2]),
+    ('POAArticle', 'application/vnd.elife.article-poa+json', [1, 2]),
+    ('VORArticle', 'application/vnd.elife.article-vor+json', [1, 2]),
+    ('ArticleHistory', 'application/vnd.elife.article-history+json', [1, 2]),
+    ('ArticleRelated', 'application/vnd.elife.article-related+json', [1, 2]),
```
thewilkybarkid: Only `application/vnd.elife.article-poa+json` and `application/vnd.elife.article-vor+json` have a version 2.

thewilkybarkid:
Requesting old versions should see a `Warning: 299 api.elifesciences.org "Deprecation: Support for version 1 will be removed` header. (If we know when, then add that too. But we probably don't.)

Try `curl -v https://api.elifesciences.org/press-packages/e6156bde -H "Accept: application/vnd.elife.press-package+json; version=1"`.

src/publisher/negotiation.py
lsh-0: cheers. that helps narrow down the targets for the middleware as well

lsh-0:
299 - Disappointed

http://mamchenkov.net/wordpress/2017/04/29/rfc-http-299-disappointed/

src/publisher/middleware.py
```diff
@@ -0,0 +1,107 @@
...
+        return [visit(row, pred, fn, coll) for row in data]
+    # unsupported type/no further matches
+    return data
+
+#
+#
+#
+
+def downgrade(content):
+    "returns v1-compliant content"
```
giorgiosironi: two key test cases to add:
- `downgrade` should be able to deal with a v2 content (because the bot-lax-adaptor will start producing only those for new articles)
- `upgrade` should be able to deal with with content that is already v2 (same reason, and its `title` field may be missing then)

lsh-0:
almost ready to go, I'm just having a problem with a fixture not regenerating. opened an issue in bot-lax-adaptor for it

lsh-0:
ok - this one is ready to merge. the bot-lax sha I'm using is older than the one in production, so I guess that means ... bot-lax needs to have it's changes merged first? or have that branch merge the changes in from master?

giorgiosironi:
I guess bot-lax-adaptor comes before this, but also api-raml possibly. I'll write up a deployment plan on the ticket.

giorgiosironi:
Merged `master`, tried to solve the conflict but check cafb479d9bfb15c7469d671dcd721059468df6c8 is correct.

giorgiosironi:
Nothing works, so I have reset to the original version that was passing. Retrying to solve the conflict.

giorgiosironi:
Finally.

giorgiosironi:
Failed badly in end2end on almost all article pages:
https://ci--alfred.elifesciences.org/job/test-lax/524/consoleFull

giorgiosironi:
```
  "message": "Uncaught PHP Exception Symfony\\Component\\Debug\\Exception\\FatalThrowableError: \"Type error: Argument 2 passed to eLife\\ApiSdk\\Model\\Block\\FigureAsset::__construct() must be of the type string, null given, called in /srv/journal/vendor/elife/api-sdk/src/Serializer/Block/FigureNormalizer.php on line 31\" at /srv/journal/vendor/elife/api-sdk/src/Model/Block/FigureAsset.php line 20"
```
Journal may be either asking for or receiving the v2, which isn't ready to deal with.

giorgiosironi:
Moving the discussion to https://elifesciences.atlassian.net/browse/ELPP-3145

PR #281, opened 2017-09-15
donotmerge - fixes a test that is calling aws

lsh-0:
locally it's calling our --prod subscriptions because of config defaults

update: there are actually more tests contacting aws while testing. looking for a way to patch boto so it never connects

### DIFF
--- a/src/publisher/aws_events.py
+++ b/src/publisher/aws_events.py
@@ -21,7 +21,7 @@ def sns_topic_arn():
     return arn
 
 def event_bus_conn():
-    sns = boto3.resource('sns')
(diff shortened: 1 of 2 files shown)


lsh-0:
I'll fix that up in this PR too.

I think I might be able to start a patcher in the `setUpClass` method of the shared super class: https://docs.python.org/3/library/unittest.html#unittest.TestCase.setUpClass

lsh-0:
huh - this is new and interesting: https://docs.python.org/3/library/unittest.html#unittest.TestCase.subTest

lsh-0:
closing. Lax is using pytest now and there is a plugin that can stop all network access

PR #292, opened 2017-10-04
Do not remove label from assets in general

giorgiosironi:
Figure-related assets must have both a label and a title in v1

https://github.com/elifesciences/api-raml/blob/develop/src/blocks/figure.v1.yaml#L36
https://github.com/elifesciences/api-raml/blob/develop/src/misc/asset-file.v1.yaml

Unfortunately this manipulation is not that safe, as we are not validating the output of `downgrade`/`upgrade`.

### DIFF
--- a/src/publisher/middleware.py
+++ b/src/publisher/middleware.py
@@ -47,11 +47,15 @@ def visit_target(content, transformer):
 def downgrade(content):
     "returns v1-compliant content"
     def transformer(item):
-        if not 'title' in item:
-            if 'label' in item:
+        if not 'title' in item: # it's a v2, as in v1 `title` is required for all assets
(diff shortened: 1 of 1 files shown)


giorgiosironi:
I can reproduce the test failure, but I don't understand the expectation.
It fails because a `label` field becomes missing:
```
    -                       'label': 'All additional files',
```

giorgiosironi:
Added comments on what is happening. Two different `downgrade` versions are likely needed to cater for `figure` blocks and `additional-assets` separately; or a parameter needs to be passed in to `transformer` to tell in which case we are.

giorgiosironi:
#293 proposes a solution; this only shows the problem

PR #293, opened 2017-10-04
Only alter label on sourceData elements.

gnott:
A quick test.


giorgiosironi:
To be tested we would need better fixtures, because the two `v1` and `v2` in:
https://github.com/elifesciences/lax/tree/develop/src/publisher/tests/fixtures/v12
do not contain figures.
A problem I'd like to verify is that when we should have to use `downgrade` on `figure`s too, as in `v1` they require `title` and the `v2` is not providing it. While I believe we don't necessarily have to use `upgrade` on `figure`s because they already required `label` for their assets?

giorgiosironi:
Update: looking at it, this should correctly run `downgrade` and `upgrade` as needed, on all `figure` `sourceData` assets; still that needs a fixture to validate it.

lsh-0:
> still that needs a fixture to validate it.

yes. @gnott , can you suggest a fixture or fixtures here?

gnott:
Looking for a good fixture, I was looking for one with sourceData elements that do not have a title and the article has some additionalFiles.

``elife-27134-v2.xml`` seems to have that.

What is doesn't have is an example of sourceData that has a label and a title. There's a number of articles with ``<supplementary-material>`` tags I can look for, if there are other cases to test - just let me know!

lsh-0:
thanks @gnott , I'll integrate that one and see how we go

lsh-0:
ok, I've replaced the fixture that was being used with the one suggested. I can get 3 of the tests to pass now but it's being a pita. Pushing what I have before I go back and check from the ground up.

lsh-0:
I think it might have some thing to do with the fixtures I'm using. For api v1 fixtures I'm using `bot-lax@master` and for v2 fixtures, I'm using `bot-lax@d0330c50ef5752d2b4609d0de20b87761ccdadd2`

is this correct?

giorgiosironi:
I regenerated the api1 and the api2 fixture with the corresponding versions and they show no change. So they seem correct.
Another direction: it's weird that there is a conflict on `src/publisher/tests/test_api_v12_temp.py` shown here.

giorgiosironi:
Merged `develop`, makes no difference on the failing tests but keeps things simple, we shouldn't have old code around in conflict with `develop`.

giorgiosironi:
The first test that fails has content in schema v1 and tries to translate it to content for schema v2. However, validation fails because (I think) this `sourceData` comes out:
```
                              'sourceData': [{'doi': '10.7554/eLife.27134.004',
                                              'filename': 'elife-27134-fig1-data1-v2.xlsx',
                                              'id': 'SD1-data',
                                              'mediaType': 'application/xlsx',
                                              'title': 'Figure 1—source '
                                                       'data 1.',
                                              'uri': 'https://cdn.elifesciences.org/articles/27134/elife-27134-fig1-data1-v2.xlsx'}],
```
This should have `label`, not `title` in v2. Hypothesis: the check in `visit_target` is too strict, trying to fiddle with it.

gnott:
Is it just how the validation error comes out that the title seems to be in two string parts,

```
'Figure 1—source '
                                                       'data 1.'
```

Does the visit function work ok when sourceData is a list of OrderedDicts?

giorgiosironi:
These commits (check out one by one as together they would create confusion) make the `test_api_v12_temp` pass; waiting for the full build.
In particular, there is a `.` inside `c6a13b31` that is missing from the schema v2 fixture. Or perhaps it's there in the v1 fixture and it shouldn't. Any ideas where it comes from @gnott?

gnott:
Dots are important. The parser strips all punctuation from the end of label values but not from title values. That would probably account for a dot here or there.

giorgiosironi:
Uhm, so the fixtures may have been correct. My commit https://github.com/elifesciences/lax/pull/293/commits/c6a13b31c667d76f9979178a3e2ed0ed0682178a should be rolledback, but the `downgrade`/`upgrade` pair needs to:
- add a end dot when downgrading?
- stripping a end dot, if present, when upgrading?

gnott:
I'm not sure - I suppose since Production requested punctuation to be stripped from label values, it would be in keeping with that request. But if converting a label to a title, we cannot add punctuation back again. It is not a validation issue so it is probably safe to just ignore the punctuation when upgrading/downgrading.

giorgiosironi:
To be fair to them, we don't want dots starting to appear in random places while we upgrade the API :)
Refining:
- do nothing when downgrading from a `label` to a `title`. Sometimes there will be missing dots and that's the price to pay.
- stripping a end dot, if present, when upgrading from a `title` to a `label`. There will never be a dot where not needed.
I don't know if this will play out well with how the fixtures are used now, because they assume they can be translated into each other, but that won't be true anymore. We may need to have instead `api1-stored`, `api1-response`, `api2-stored`, `api2-response`?

gnott:
Actually, as a side issue, when I introduced in the parser the prefer_label flag for adding title/label/caption values, I didn't strip punctuation from a title that gets converted to a label. I will issue a small update to elife-tools to reflect that and make sure it is test backed. At least that would make the logic there to produce the best output.

For punctuation stripping, it actually uses rstrip_punctuation() (https://github.com/elifesciences/elife-tools/blob/develop/elifetools/utils.py#L461) which strips . or :   It can be used here if you want by
from elifetools.utils import rstrip_puncutation

src/publisher/tests/test_api_v12_temp.py
```diff
@@ -76,13 +81,24 @@ def test_all(self):
...
+
                     self.publish_ajson(content)
                     actual_resp = self.c.get(reverse('v2:article', kwargs={'id': self.msid}), HTTP_ACCEPT=request)
                     self.assertEqual(actual_resp.status_code, 200)
                     # response is valid
                     self.assertTrue(utils.validate(actual_resp.json(), schema_idx[rskey]))
                     # response is equal to what we're expecting
-                    self.assertEqual(self.load_ajson2(actual_resp.json()), response)
+                    actual_json = self.load_ajson2(actual_resp.json())
+                    # slightly more isolated error messages
```
lsh-0: if this was beneficial, I might turn it into a function `compare_ajson`

giorgiosironi:
Only helped a bit, but more importantly it would provide a center of gravity for further improvements to the comparison algorithm :+1: 
@gnott waiting on an additional `elifetools` update then?

giorgiosironi:
Updated with new bot-lax-adaptor, let's see what happens.

PR #296, opened 2017-10-18
Bugfix deadlock debug branch

lsh-0:


lsh-0:
hey @seanwiseman , I'm not sure what your workload or focus is like at the moment, but I included you as a reviewer. If you have any experience with python's `multiprocessing` I'd love some feedback.

.test.sh
```diff
@@ -20,6 +20,7 @@ find src/ -name '*.pyc' -delete
 # called by test.sh
 rm -f build/junit.xml
 #./src/manage.py test --testrunner=green.djangorunner.DjangoRunner "$@"
+MULTIPROCESSING=1 # testing management commands that require a queue shared between processes
```
giorgiosironi: I think you need `export` here? https://stackoverflow.com/questions/1158091/defining-a-variable-with-or-without-export

giorgiosironi: since this is a very specific environment variable, it could be prefixed with `LAX_` to avoid clashing with other projects/libraries/systems/etc

seanwiseman: From what you have done it all looks perfectly functional to me. I have nothing to add to what you have done but to give you another view on handling similar tasks I have provided a method I have used a lot.

I have used multiprocessing a fair amount in previous projects, the most common scenario was having to ingest serialized data via socket connections (such as images) from multiple remote clients, which would need to be post processed and stored.

All I was interested in was dumping tasks to a queue and having my pool of workers pick them off 'and' tell me once done. This way there was little need to check the queue for state often as you could hook into the processing telling you its finished a queue task via the `queue.task_done()` method which is available when using a `JoinableQueue()`. 

Note in this scenario I did not have to share data between processes, as there needs were independent and did not require a manager.

Below is just some snippets from a previous project for clarity. I init the queue and workers in the projects main class, I sub class the `Process` class and add some functionality to its `run` method which gets executed when you call `Process.start()`.

```
def _init_image_processing(self):
    """Init image processing components"""

    # set up queue
    self.image_processing_queue = multiprocessing.JoinableQueue()

    # set up image workers
    self.num_workers = multiprocessing.cpu_count()  # * 2
    self.image_workers = [
        ImageWorker(
            task_queue=self.image_processing_queue,
            screen_shots_end_point=self.screen_shots_end_point,
            name="image_worker_{0}".format(worker),
            config=self.config
        ) for worker in range(self.num_workers)
    ]

    for worker in self.image_workers:
        worker.start()
```

```
class ImageWorker(multiprocessing.Process):
    def __init__(self, task_queue=None, screen_shots_end_point=None, name="worker", config=None):
        multiprocessing.Process.__init__(self)
        self.task_queue = task_queue
        self.screen_shots_end_point = screen_shots_end_point
        self.config = config
        self._process_img_func = None
        self.name = name

    def run(self):

        self.img_mgr = ImageManager(
            api_end_point=self.screen_shots_end_point,
            config=self.config
        )
        self.img_mgr.api_data_processor.start()

        # setup image processing func based on config
        if self.json_to_bool(self.config['screenShots'].get('s3Storage', False)):
            self._process_img_func = self.img_mgr.process_image_data
        else:
            self._process_img_func = self.img_mgr.process_image_data_local

        while True:
            image_data = self.task_queue.get()

            if image_data is None:
                self.task_queue.task_done()
                break

            self._process_img_func(image_data)
            self.task_queue.task_done()
```

giorgiosironi: Note: we use this pick-work-and-commit-as-done pattern with SQS, as that's the API those queue expose https://github.com/elifesciences/bot-lax-adaptor/blob/develop/src/sqs_adaptor.py#L39-L50

seanwiseman: Ah OK I see, thank you.

lsh-0: thanks @seanwiseman , that was insightful. my decision to go with a manager-less SimpleQueue in `aws_events.py` was naive. I thought it would do some crazy python magic behind the scenes to deal with multiple producers when multiprocessing. And then I forgot it was there. It also doesn't behave like a regular queue or multiprocessing.queue. 

@giorgiosironi , I revisited your implementation and still preferred my version, but they both suffered from the same problem. I also wanted to work through the problem rather than just discard/re-architect code.

lsh-0:
feel free to merge this if it hasn't finished testing by the time I sign off

lsh-0:
phew - took a while, but it's in now.

.test.sh
giorgiosironi: On the other implementation: they have different designs for the interface but they have the same bug because the underlying implementation was based on `SimpleQueue` in both cases

PR #301, opened 2017-10-24
test scripts shifted to use .test-green.sh over .test.sh

lsh-0:
looks like green is counting fewer files than the coverage test runner and is reporting a 75% pass rather than coverage's own 88%

### DIFF
--- a/.cc-check.sh
+++ b/.cc-check.sh
@@ -17,7 +17,7 @@ xenon src/ \
 	--max-absolute B \
 	--max-modules A \
 	--max-average A \
-	--exclude '*eif_ingestor.py,*ajson_ingestor.py,*ingest.py' || {
+	--exclude '*ajson_ingestor.py,*ingest.py' || {
 
     echo "use 'radon cc path/to/file.py' for more detail" > /dev/stderr
     exit 1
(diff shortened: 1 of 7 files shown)


giorgiosironi:
https://github.com/elifesciences/lax/blob/54a492e8a68003e2fdcaf8aa7a5d3827329767eb/src/core/settings.py#L134 seems to say the `junit.xml` file for test results is produced by Django no matter what the runner is, so it should still be there?

giorgiosironi:
If you run into unsolvable problems, another runner we use for parallelization is pytest with the https://pypi.python.org/pypi/pytest-xdist extension. Don't know how it works with the coverage as in end2end tests (where it is used) it's meaningless to measure.

lsh-0:
> seems to say the junit.xml file for test results is produced by Django no matter what the runner is, so it should still be there?

yes, I thought I may have removed it years ago when you started. I can see the `./build/junit.xml` file now. just look at all that wonderful timing data ...

giorgiosironi:
```
LAX_MULTIPROCESSING=1 ./src/manage.py test publisher.tests.test_utils
```
did produce a `build/junit.xml` file for me locally, but
```
GREEN_CONFIG=.green LAX_MULTIPROCESSING=1 ./src/manage.py test publisher.tests.test_utils --testrunner=green.djangorunner.DjangoRunner
```
didn't, leaving `build/` empty.

PR #316, opened 2017-11-24
Feat: improve validation failed message

lsh-0:
original intent of PR was to disable validating against v1 content, but that is still being used until all clients are upgraded to requesting v2 content. instead, I've improved the error message.


giorgiosironi:
I wonder where we need to validate against v1 now:
- all content is v2 (has been backfilled)
- all new ingests are v2 (bot-lax-adaptor and elife-tools are in their recent versions)
- v1 is produced by downgrading by running https://github.com/elifesciences/lax/blob/64fd2c4e8b6faae7eb80af19851ab8b0a6fbe202/src/publisher/middleware.py#L47-L58

so unless we run validation after `downgrade`, we should not need to validate against v1 at all?

lsh-0:
I disabled the v1 poa/vor schemas, but it's used in testing. The test is that valid content is produced after a v2 downgrade to v1: https://github.com/elifesciences/lax/blob/develop/src/publisher/tests/test_api_v12_temp.py#L47

once all clients have been upgraded to request v2 content (or some arbitrary cut off date) we can properly disable v1 schema

PR #327, opened 2017-12-15
[WIP] Add GraphQL functionality

seanwiseman:
A small exploration of seeing what a GraphQL interface looks like through Lax.

This is only meant for exploring ideas and holds no weight in terms of schema design or future data modeling decisions.

### DIFF
--- a/src/core/schema.py
+++ b/src/core/schema.py
@@ -1,0 +1,11 @@
+import graphene
+
+from publisher.schema import Query as PublisherQuery
+
+
+class RootQuery(PublisherQuery, graphene.ObjectType):
+    # This class will inherit from multiple Queries
+    pass
+
+
+schema = graphene.Schema(query=RootQuery, mutation=None)
--- a/requirements.txt
+++ b/requirements.txt
@@ -14,6 +14,7 @@ djangorestframework==3.6.3
 et3==1.5
 # versions of green > 2.9.0 conflict with kids.cache
 # both are doing command line parsing during installation
+graphene-django==2.0.0
 green==2.9.0
 joblib==0.11
 jsonschema==2.6.0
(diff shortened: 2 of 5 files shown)


lsh-0:
could you add some example usage in the form of tests, please? 
REST is pretty straightforward, there's even a gui built in, but if we do investigate Graphql seriously in future we should be able to point to lax and go "here is a python+django implementation, here is how to use it, this is what you can expect as a response", etc.

seanwiseman:
I have added tests along with the query strings required to generate the expected response value(s). Please note some of these are obvious (find an `Article` by `manuscript_id`) and some I just made up (find `ArticleEvent`'s by `article__manuscript_id`) to show what could be done, they might be useless in practice.

These should give you an idea of the basic query structure now exposed, this can be extended quite some bit still but it can do a fair amount in its current state. 

Along with the tests, the endpoint for the GraphQL queries ( http://host:port/graphql ) will open the interactive [GraphiQL](https://github.com/graphql/graphiql) IDE. From here you can play with the data using the query examples in the tests and also use the documentation explorer on the right of the page.

lsh-0:
cheers, @seanwiseman, I'll take a closer look. Lax contains sensitive information (unpublished articles, mostly) so I may introduce a feature flag that has graphql disabled by default, I just need to see what it's doing for myself.

src/core/schema.py
```diff
@@ -0,0 +1,11 @@
+import graphene
```
giorgiosironi: this module (`schema`) seems to be GraphQL-specific but its naming doesn't reflect that, so it could be confused with the relational schema

seanwiseman: I was following the naming convention used in [graphene-python](http://graphene-python.org/), but yes I agree, its very broad and could be named differently.

lsh-0: I had the same thoughts, lets change it.

lsh-0:
(those tests were very helpful in getting me started btw, cheers)

PR #328, opened 2017-12-18
article-json hash checking

lsh-0:
initial commit to handle ingestion of identical article-json.
* `fragment_logic.set_article_json` now accepts a `check_hash` argument that will compare old hashes with new and fail with an `Identical` runtime error.
* hash checking has been deliberately disabled for several edge cases, it's intent is really to cut down on article update events being emitted during a backfill.
* lax now preserves the ordering of all `JSONFields`
* lax now has a `rehash` command that will regenerate the hashes for all of the stored article-json. takes about 2mins.
* `ingest` command now returns a successful `INGESTED` response if `Identical` error raised during ingest. 

event count checking test here: https://github.com/elifesciences/lax/pull/328/files#diff-071f44566bdd9c069b8c23d7e54eb1d0R271

depends on:
1. ~https://github.com/elifesciences/bot-lax-adaptor/pull/377~ merged


src/publisher/tests/test_cmd_ingest.py
```diff
@@ -204,6 +205,14 @@ def test_multiple_ingest_from_cli(self):
             # publish is called once per article (not article version)
             self.assertEqual(expected_art_count, mock.publish.call_count)
 
+        # tweak article data to avoid failing hashcheck
```
lsh-0: note to self: I could probably just skitch the hash in the `av`

src/publisher/fragment_logic.py
```diff
@@ -181,24 +187,39 @@ def merge_if_valid(av, quiet=True):
...
     """updates the article with the result of the merge operation.
     if the result of the merge was valid, the merged result will be saved.
     if invalid, the ArticleVersion instance's article-json will be unset.
     if invalid and quiet=False, a ValidationError will be raised"""
-    log_context = {'article-version': av, 'quiet': quiet}
+    log_context = {'article-version': av, 'quiet': quiet, 'hash_check': hash_check}
     result = merge_if_valid(av, quiet=quiet)
+    newhash, oldhash = hash_ajson(result), av.article_json_hash
+    if hash_check and oldhash == newhash:
+        raise Identical("article data is identical to the article data already stored", av, newhash)
```
giorgiosironi: An exception for a supported flow (idempotence of ingest) is suspicious, can it be substituted with an early `return`?

src/publisher/fragment_logic.py
```diff
@@ -181,24 +187,39 @@ def merge_if_valid(av, quiet=True):
...
     av.article_json_v1 = result
     av.article_json_v1_snippet = extract_snippet(result)
+    av.article_json_hash = newhash
     av.save()
     if not result:
         msg = "this article failed to merge it's fragments into a valid result. Any article-json previously set for this version of the article has been removed. This article cannot be published in it's current state."
-        LOG.warn(msg, extra=log_context)
+        LOG.warn(msg, extra=log_context) # TODO: consider upgrading to 'critical' ?
```
giorgiosironi: indeed, it seems like a very serious error potentially corrupting an article

src/publisher/fragment_logic.py
```diff
@@ -181,24 +187,39 @@ def merge_if_valid(av, quiet=True):
...
+        self.hashval = hashval
+
+# TODO: rename `quiet` to `valid_check` or similar.
+def set_article_json(av, quiet, hash_check=True):
     """updates the article with the result of the merge operation.
     if the result of the merge was valid, the merged result will be saved.
     if invalid, the ArticleVersion instance's article-json will be unset.
     if invalid and quiet=False, a ValidationError will be raised"""
-    log_context = {'article-version': av, 'quiet': quiet}
+    log_context = {'article-version': av, 'quiet': quiet, 'hash_check': hash_check}
```
giorgiosironi: I would check that the changes to this `fragment_logic` module work well with the POST fragment API https://github.com/elifesciences/lax/blob/fa2b4a55a7475dafec83b17d88e48bb9b5e3fe87/src/publisher/api_v2_views.py#L133-L167
For example, there is the additional `Identical` exception to deal with and possibly more. It's quite a similar flow however, an event should be sent if the POST of the fragment updates the article, but not if the hash doesn't change, etc.

src/publisher/management/commands/ingest.py
```diff
@@ -7,19 +7,20 @@
 
 """
 from collections import OrderedDict
```
giorgiosironi: a feedback from Digirati's and our containerized usage has been: sooner or later we could try to make the ingest called by bot-lax-adaptor available through HTTP rather (or in substitution) of them relying on being on the same filesystem and creating a new Lax process. This would decouple them and let them run in different containers. We don't want to do decisions here that makes this difficult, at the moment this command seems self-contained and could be restructured into an API if needed. :+1:

src/publisher/fragment_logic.py
lsh-0: I can raise the exception here or I can return a different value and raise the exception there, but I need to raise something to get the transaction to rollback. Using exceptions as flow control isn't a large sin and this is a delicate area of the codebase - I'd prefer to follow existing patterns and raise the exception.

src/publisher/management/commands/ingest.py
lsh-0: we'll cross that bridge when we come to it.

src/publisher/fragment_logic.py
lsh-0: good catch, I'll investigate

lsh-0: fixed up. actually found a case that wasn't handled and fixed that up too.

src/publisher/tests/test_api_v2_views.py
```diff
@@ -591,14 +534,91 @@ def test_add_bad_fragment(self):
...
+    def test_add_invalid_fragment(self):
+        "request with fragment that would fail the ArticleFragment schema is refused"
+        self.assertEqual(models.ArticleFragment.objects.count(), 1) # xml->json
+        fragment = {} # empty
+        url = reverse('v2:article-fragment', kwargs={'art_id': self.msid, 'fragment_id': 'test-frag'})
+        resp = self.ac.post(url, json.dumps(fragment), content_type="application/json")
+        self.assertEqual(models.ArticleFragment.objects.count(), 1) # 'xml->json'
+        self.assertEqual(resp.status_code, 400) # bad client request
+
+    def test_add_fragment_causes_no_change(self):
```
giorgiosironi: :+1:

lsh-0:
bot-lax dep has been merged in. I'll wait until it's in production before merging this.

PR #332, opened 2018-01-12
multiprocessing.Manager instance available only with LAX_MULTIPROCESSING

lsh-0:
https://elifesciences.atlassian.net/browse/ELPP-3295

I don't really know how uwsgi is doing it's process forking and it's interaction with the creation of a multiprocessing.Manager instance and how these relates to the two different errors we're seeing (cannot join thread, cannot join child process), but there is no reason for a Manager instance to be created when the lax process is started by uwsgi. it's used so the command line ingest and aws event sending can converse.

PR #334, opened 2018-01-17
LAX_MULTIPROCESSING var must be set when ingesting articles in parallel

lsh-0:
related to https://github.com/elifesciences/lax/pull/332

### DIFF
--- a/src/publisher/management/commands/ingest.py
+++ b/src/publisher/management/commands/ingest.py
@@ -10,7 +10,7 @@ from collections import OrderedDict
 import os, io, re, sys, json, argparse, time
 from publisher import ajson_ingestor, utils, codes, aws_events
 from publisher.aws_events import START, STOP
-from publisher.utils import lfilter, formatted_traceback as ftb, ensure
+from publisher.utils import lfilter, formatted_traceback as ftb
 from publisher.ajson_ingestor import StateError
 import logging
 from joblib import Parallel, delayed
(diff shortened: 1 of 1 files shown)


giorgiosironi:
Or, perhaps it shouldn't be trying to use the parallel mode during normal ingests.

giorgiosironi:
Transformed the assertion into a log so that this can be merged without stopping the pipeline.

giorgiosironi:
Long build, `lax--ci` ate all its credits with a couple of builds. Perhaps it should be resized since it's a `t2.small` running a 15-minute test suite.

lsh-0:
I wasn't clear on where bot-lax was calling the ingest script in the formula and failing, so I introduced the error to see if I could get a failed state. gave your  PR a :+1:  

> 15-minute test suite

do we have metrics on which tests are taking the longest? I feel tests shouldn't take more than 10mins.

running tests with green+sqlite is very fast, but it may miss some db field related issues (constraint checks) and running with green+psql introduces weird locking problems.

giorgiosironi:
> do we have metrics on which tests are taking the longest? I feel tests shouldn't take more than 10mins.

https://alfred.elifesciences.org/job/test-lax/lastCompletedBuild/testReport/
Ignore `(root)` as it contains end2end tests which are run in parallel. The 15 minutes figure includes deployment time.

lsh-0:
hmm - definitely room for some optimisation, but it's only just hovering around the 10min goal so I'll leave it for now

PR #349, opened 2018-01-30
added the 'privileged' database to settings.

lsh-0:
this uses the root db credentials in the app.cfg file to run migrations outside of development.

not sure if this is the best route or if granting the app user db modification privileges is, but I was getting frustrated figuring out what specific set of additional GRANTs the app user needed in postgresql. 
this way at least it need not be a root user just simply another user but with elevated privileges

depends on https://github.com/elifesciences/lax-formula/pull/23

### DIFF
--- a/elife.cfg
+++ b/elife.cfg
@@ -20,5 +20,7 @@ name: lax.sqlite3
 engine: django.db.backends.sqlite3
 user: 
 password:
+root-user: 
+root-password: 
 host: 
-port: 
+port:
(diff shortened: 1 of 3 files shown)


giorgiosironi:
Not in favor of allowing the application to log in with the `root` user - seems like too much power e.g. creating or removing users?

I think it would be fine to add another non-root user with elevated permissions, but:
- if it's in the same configuration file as the main one, why separating it at all?
- this is not scalable to generic Python applications, which would *all* need to be configured with two users, increasing their complexity

I'll go through the failed builds to propose alternatives. Another point for zero-maintenance RDS :)

lsh-0:
ticket here: https://elifesciences.atlassian.net/browse/ELPP-3417

PR #350, opened 2018-01-31
[wip] adds dc:date.earliest element to report rss feeds

lsh-0:
ticket: https://elifesciences.atlassian.net/browse/ELPP-3409

lax was once going to be a mighty data collection and reporting tool but it's purpose has evolved into being something of a proxy between the production realm where everything is xml and the post-production realm where everything is json. 

these rss feeds originally duplicated the content found on the original elife websites and are now being used by production services outside of elife who cannot (for some reason) transition from xml to csv/json/whatever.

I would love to just cut all this nasty OO code out. Django's syndication feeds are not great and Observer opted to use a different library for generating them.


giorgiosironi:
Not able to comment much over Django, but then is there a plan to sooner or later port these into observer or should they just live here until they are decommissioned years (sooner or later too)?

lsh-0:
I think this is a conversation we need to have with Melissa/Production/Exeter. They're only using rss because for the longest time that was the only structured data they could get out of the website. And RSS is a PITA to model certain attributes.

PR #352, opened 2018-02-13
lint script will now fail on certain warnings.

lsh-0:
first warning we'll be failing on is shadowing builtins

### DIFF
--- a/.lint.sh
+++ b/.lint.sh
@@ -14,8 +14,8 @@ pyflakes ./src/
 
 echo "pylint"
 pylint -E ./src/publisher/** --load-plugins=pylint_django --disable=E1103 2> /dev/null
-# specific warning we're interested in, comma separated no spaces
-# presence of these warnings are a failure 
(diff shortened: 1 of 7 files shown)


lsh-0:
not sure what the state of proofreader is. After failing to find anything that checked for built in shadowing, I told Sean if he found one I'd make all the code changes.

seanwiseman:
The state of [proofreader-python](https://github.com/elifesciences/proofreader-python) is it is currently being used by 4 projects in place of installing and calling `pylint` and `flake8` separately. It uses defaults for both defined in the package but can be overridden if need be in the normal way. (dependency license checking is also coming soon)

An example: [profiles](https://github.com/elifesciences/profiles/blob/develop/project_tests.sh)

It would currently fail on builtin shadowing detection (as we would want it to in this instance) as the default `pylint` config is not ignoring warnings. If you do want to ignore some warnings and allow others however then you need to do what has been done in this PR. This then falls under the caller overriding the config.

PR #363, opened 2018-03-14
Customize TTL of responses

giorgiosironi:
This is useful in testing environments as we don't have to wait 5 minutes before a silent correction is propagated to Journal after cache expiration.
e.g. https://github.com/elifesciences/elife-spectrum/pull/108

### DIFF
--- a/src/core/middleware.py
+++ b/src/core/middleware.py
@@ -51,8 +51,8 @@ class DownstreamCaching(object):
     def __call__(self, request):
         public_headers = {
             'public': True,
-            'max-age': 60 * 5, # 5 minutes, 300 seconds
-            'stale-while-revalidate': 60 * 5, # 5 minutes, 300 seconds
+            'max-age': settings.CACHE_HEADERS_TTL,
+            'stale-while-revalidate': settings.CACHE_HEADERS_TTL,
             'stale-if-error': (60 * 60) * 24, # 1 day, 86400 seconds
         }
         private_headers = {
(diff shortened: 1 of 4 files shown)


src/core/middleware.py
```diff
@@ -51,8 +51,8 @@ def __init__(self, get_response):
     def __call__(self, request):
         public_headers = {
             'public': True,
-            'max-age': 60 * 5, # 5 minutes, 300 seconds
-            'stale-while-revalidate': 60 * 5, # 5 minutes, 300 seconds
+            'max-age': settings.CACHE_HEADERS_TTL,
+            'stale-while-revalidate': settings.CACHE_HEADERS_TTL,
```
thewilkybarkid: We might want to set this to 0 in `end2end`, should force the refresh before looking at the page rather than after.

giorgiosironi: Is it totally equivalent to a page that gets cached and then expire very soon? End2end knows how to deal with cached pages by continuing to poll for a new result.

thewilkybarkid: Yeah, might just be slightly quicker (avoid some polling, but the request from Journal happens when producing the page - depends how often it polls).

src/core/tests.py
```diff
@@ -64,3 +65,14 @@ def test_cache_headers_not_in_response(self):
         resp = self.c.get(self.url)
         for header in cases:
             self.assertFalse(resp.has_header(header), "header %r present in response" % header)
+
+    @patch('core.middleware.settings')
```
lsh-0: Django has this to patch settings during testing: https://docs.djangoproject.com/en/2.0/topics/testing/tools/#django.test.override_settings

but I think the intent here is just as clear.

PR #370, opened 2018-03-27
Generate friendly error path string

seanwiseman:
When an error is not in a `oneOf` check (you can get the full error path):

`Issue: an invalid funderId`

**Current message:** 
```validation error: '0000' does not match '^10[.][0-9]{4,}[^\\s"/<>]*/[^\\s"]+$'```

**With changes:**
```
validation error: '0000' does not match '^10[.][0-9]{4,}[^\\s"/<>]*/[^\\s"]+$' 

*path to field *: funding >> awards >> 2 >> source >> funderId = 0000
```

When an error is in a `oneOf` check (you can only get so far):

`Issue: an invalid author orcid`

**Current message:**

```validation error: {'type': 'person', 'name': {……...rest of the data... is not valid under any of the given schemas``` 

**With changes:**

```
validation error: {'type': 'person', 'name': {……...rest of the data... is not valid under any of the given schemas 

*path to field *: authors >> 4 = {'type': 'person', 'name': { ……...rest of the data...
```

I tried playing with `best_match(Draft4Validator(schema).iter_errors(struct))`.

You can use it (in a limited way) if you just wanted it to traverse into oneOfs for you by calling best_match([e]), but it turned out to not always give the required error message, as you know you don’t know which `oneOf` you were targeting in the first place, it failing on one might be perfectly fine because that was not the expected structure. It could have some legs but would get quite involved.

You can obviously get access to the full trace but the target is a short summary/pointer towards problem data without having to do too much digging.

### DIFF
--- a/src/publisher/tests/test_ajson_ingest.py
+++ b/src/publisher/tests/test_ajson_ingest.py
@@ -435,7 +435,8 @@ class Publish(BaseCase):
             ajson_ingestor.ingest_publish(self.ajson)
         except StateError as err:
             self.assertEqual(err.code, codes.INVALID)
-            self.assertEqual(err.trace.strip()[:15], "'' is too short")
+            # self.assertEqual(err.trace.strip()[:15], "'' is too short")
+            self.assertEqual(err.trace.strip()[:23], "title = '' is too short")
 
 class IngestPublish(BaseCase):
     def setUp(self):
--- a/src/publisher/tests/test_utils.py
+++ b/src/publisher/tests/test_utils.py
@@ -44,9 +44,7 @@ class Errors(base.BaseCase):
 
             # non-deterministic ordering of errors! hooray
 
-            expected1 = """'status' is a required property
-
-Failed validating 'required' in schema['allOf'][0]['allOf'][0]:"""
+            expected1 = """'status' is a required property"""
 
             expected2 = """None is not of type 'string'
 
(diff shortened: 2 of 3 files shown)


seanwiseman:
I have not updated the test that is failing as its referring to error string contents (of which is different due to changes) and it is unlikely to stay in this exact format.

seanwiseman:
@lsh-0 this is more exploratory at the moment so your input is more than welcome.

lsh-0:
It's a good start and insightful. 

I think we could craft some transformations to an otherwise valid article that would throw specific validation errors. We could then check the generated error against this invalidation.

That spreadsheet Melissa/Production have is a goldmine, we should use it to start a set of testcases with. For example, the sheet has "Lax error" and "Meaning" fields where those guys are translating between the two. The first row is:

> validation error: 'statement' is a required property

and their interpretation is

> Funding statement had been deleted entirely

for that class of errors ('x' is a required property), we may be able to emit something close to their meaning, like `path.to.funding.statement: required but missing.`. 

The path would give some context into where in the article the problem lies and might even serve as an id to override specific general errors with more specific ones. 

For example, say an author is missing a last name (contrived and we also don't have any such field explicitly). You would get the above error `'surname' is a required property` which we could introspect and come up with a path `authors.3.name.surname`. The general error would be `authors.3.name.surname: required but missing` (or whatever) but we know that paths that look like `authors.N.name.surname` with the error `required` are so common we could replace the general case with the more humane `author 'Christopher' is missing a surname`. 

Or, `funding statement has been deleted entirely`.

giorgiosironi:
> I think we could craft some transformations to an otherwise valid article that would throw specific validation errors. We could then check the generated error against this invalidation.

You mean as a one-off or a build step or a runtime step?

I like the idea that we can, best-effort, decorate some errors with more specific informations or details. Apart from parsing/massaging the error message that comes out of the library, I don't see easy ways to do it however.

lsh-0:
> You mean as a one-off or a build step or a runtime step?

sorry if it was unclear, I meant in terms of getting some tests set up against actual known errors with obscure messages and the ideal versions.

seanwiseman:
I have changed the path format and added a test that shows an error response from a `funderId` being invalid.

message:
`funding.awards.1.source.funderId = '0000' does not match '^10[.][0-9]{4,}[^\\s"/<>]*/[^\\s"]+$'`

I am not sure how close we can get to the ideal message versions suggested without some further parsing magic and further error handling logic 🤔, this is however a small improvement if this is the kind of message that ends up viewable by the production team (or whoever has access).

lsh-0:
I think the dotted path is definitely a step up. Good work

I'll take a look today at that spreadsheet and your work and see if I can't
tackle some of the thornier problems, like multi schema failures.

I wonder if emitting the parent structure in that path ('source') as part
of the message would be helpful? I'll toy with that as well

seanwiseman:
@lsh-0 did you get round to playing with this?

lsh-0:
I didn't actually, remastering our minions went from being a 1hr job to a three day job. It's still on my radar though.

seanwiseman:
Merging this change in as it does not pose any detrimental side effects and could be of use with some current errors that get raised. Is also easily reversible.

Future work in this area can be picked up when other priorities are not taking precedence.

lsh-0:
No problems, and apologies for being so slow to getting around to it.

PR #374, opened 2018-05-04
validation failures are logged as warnings when caught by  ingest 

lsh-0:
another instance in lax where validation failures are being reported as errors rather than as a warning that Loggly will then send to us.

validation failures are an ordinary part of the production workflow


lsh-0:
> If the intent is not to receive emails on validation failures, because they are a normal part of the workflow, 

precisely. I downgraded a similar one last year but this one has been hanging around for a while. I thought I'd handle it while I was handling the elife-metrics error emails.

PR #387, opened 2018-07-02
fix: capped error messages

lsh-0:
validation failures with more than N (10) errors are given a slightly different error messages and the output is capped at 10.

same for iterating through individual error possibilities but we don't tweak those messages


seanwiseman:
> cheers, @seanwiseman 

Not sure of the reason for this, I was not working yesterday so was therefore not reviewing any PRs.

lsh-0:
oh sorry, I thought you had reviewed it. I could have been looking at something else at the time

seanwiseman:
Ok np.

PR #391, opened 2018-07-06
downgrades a schema validation warning from ERROR to WARN

lsh-0:


lsh-0:
absolutely. this is the third and (hopefully) final validation error that is getting reported to techalerts@...

PR #398, opened 2018-08-13
caps sub-errors in validation errors when multiple schemas involved

lsh-0:


lsh-0:
in the case shown to me there were 16 and possibly more sub-errors going off the page. They're not terribly useful either, but with such a large schema as the article-json one then things are going to get cryptic

lsh-0:
giving build a kick, Alfred was having a bad day when this failed

PR #401, opened 2018-09-07
relaxes version of python to use

lsh-0:
`python3` in 18.04 is `python 3.6`, so the 16.04 branch of the formula builds (python 3.5) but not an 18.04 one

fyi @giorgiosironi

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -4,15 +4,29 @@ echo "[-] install.sh"
 
 . download-api-raml.sh
 
-if [ ! -e "venv/bin/python3" ]; then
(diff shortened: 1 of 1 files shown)


lsh-0:
depends on: https://github.com/elifesciences/builder-base-formula/pull/175

lsh-0:
merging this will cause lax--prod to re-create the venv on next highstate and go from python 3.5 to 3.6

opened up to review.

lsh-0:
apparently you're still on holiday, merging

PR #406, opened 2018-10-08
Feature: dep bump

lsh-0:


canary.sh
```diff
@@ -21,7 +15,4 @@ pip-review --pre # preview the upgrades
 echo "[any key to continue ...]"
 read -p "$*"
 pip-review --auto --pre # update everything
-
-# run the tests
-python src/manage.py migrate
-./src/manage.py test src/
+pip freeze > new-requirements.txt
```
giorgiosironi: this script could be run in a pipeline (using `new-requirements.txt` to overwrite `requirements.txt`?) Or does it need human cleanup and would need to use `pipenv` instead?

lsh-0: I don't like pipenv. I'd rather change our usage of requirements.txt file to be the complete output of `pip freeze` and capture the direct requirements elsewhere.

the original intention way back when was for this canary script to be automated, it probably still could, but there some issues:
* Django upgrades are involved and not safe unless it's just the patch version
* some libraries don't mark alpha/beta/unstable releases as such and we get *weird* errors. `coverage` and `pylint` and `asteroid` are classic examples of this.

if we could specify some libraries to only upgrade themselves within the patch or minor versions then the canary script would probably fail less often.

perhaps `requirements.txt`:
```
django<2.0
coverage<5.1
...
```
and `requirements.frozen.txt`:
```
django==1.11.16
dep-of-django==1.2.3
coverage==5.0.9
dep-of-coverage==2.3.4
dep-of-dep-of-coverage==1.3.5
```

giorgiosironi: TIL that pipenv is only recommended for applications not libraries, which is not a small world to be excluded; by the way long-term decisions here are to be the responsibility of [a new hire](https://elifesciences.org/jobs/11fba93e/senior-developer-python-devops-aws-open-source)

lsh-0: I would love to give up responsibility and decision making to a new hire, but Sean came and left so quickly I'm hesitant to do so now. We do have a number of 'orphaned' python projects that need leadership, most others like elife-bot, elifetools, lax, metrics, are stable and have owners and wouldn't really benefit from new tooling. I'm not present for the interviews, but perhaps you could ask them how they would handle that class of project?

I don't see pipenv recommended anywhere except in gushing medium articles. The author is a little self-aggrandizing too and gives off a weird vibe. I've tried to incorporate it into my workflow but had too much resistance.

lsh-0: (what does TIL mean?)

giorgiosironi: (Today I Learned)

giorgiosironi: (the main author is [the `requests` guy](https://github.com/kennethreitz))

>  you could ask them how they would handle that class of project?

@BlueReZZ often brings up "leading on all things Python" as the goal, but I'll leave this vision to him to explain; also probably moving this discussion somewhere private

lsh-0: > (the main author is the requests guy)

yes, that gave me pause when I first found out. I think he should stick to opt-in libraries rather than replacing toolchains.

PR #409, opened 2018-10-29
http error responses now contain no-caching directives

lsh-0:


src/core/middleware.py
```diff
@@ -49,22 +49,29 @@ def __init__(self, get_response):
...
+            'no-cache': True,
+            'no-store': True
+        }
 
         authenticated = request.META[settings.KONG_AUTH_HEADER]
-        headers = public_headers if not authenticated else private_headers
+        headers = public_cache_values if not authenticated else private_cache_values
 
         response = self.get_response(request)
+        if response.status_code > 399:
```
giorgiosironi: larger scope than just 404, but it makes sense to me to avoid persisting temporary errors for minutes

src/core/middleware.py
```diff
@@ -49,22 +49,29 @@ def __init__(self, get_response):
         self.get_response = get_response
 
     def __call__(self, request):
-        public_headers = {
+        public_cache_values = {
```
thewilkybarkid: 'directive' is the correct term.

src/core/middleware.py
```diff
@@ -49,22 +49,29 @@ def __init__(self, get_response):
...
         }
-        private_headers = {
+        private_cache_values = {
             'private': True,
             'max-age': 0, # seconds
             'must-revalidate': True,
         }
+        error_cache_values = {
+            'must-revalidate': True,
+            'no-cache': True,
```
thewilkybarkid: Following https://stackoverflow.com/a/2068407, `no-cache` is redundant (but harmless).

PR #412, opened 2018-11-22
Feature: forced INGEST requests on v1 articles now sets pubdate

lsh-0:
This turned out to be less of a bug and more of a deliberate, thoroughly tested design decision. 

The intent was that if you wanted to change the publication date for an already-published article, you would send a forced PUBLISH event for that article. 

Unfortunately this behaviour was forgotten about, including by me, as a forced INGEST ('silent correction') is assumed to update the article data without exception. In the design, this is true, but with the pubdate being the only exception ;)

So for better or worse, this introduces a feature/fixes a bug where silent corrections can alter the publication date.

Caveat: To change the original pubdate of an article the silent correction must be issued for the v1 article.  If the bad date is also present in subsequent versions of the xml, they need silent corrections as well for consistency but are generally not used (but *are* available as the raw unprocessed fragment data sent to lax).

### DIFF
--- a/src/publisher/ajson_ingestor.py
+++ b/src/publisher/ajson_ingestor.py
@@ -69,6 +69,8 @@ def _ingest_objects(data, create, update, force, log_context):
 
         av_struct = render.render_item(ARTICLE_VERSION, data['article'])
         # this is an INGEST event and *not* a PUBLISH event. we don't touch the date published.
+        # 2018-11-22: not strictly true anymore, forced INGEST events are expected to change pubdates
+        # we'll handle that further down in _ingest though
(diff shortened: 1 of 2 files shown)


lsh-0:
> Can the pub date still be changed in normal, non-force ingest runs, before the version 1 of the article is published?

yes. The pubdate is only 'set' when a PUBLISH event is sent. After that, a forced INGEST or forced PUBLISH event will be required to re-set the pubdate.

> but I thought we had a previous example where the first pub date supplied did not change even though it was changed in the XML.

Quite possibly this issue or related to it. If you can recall a specific ticket I'm happy to look at it in regards to this

PR #414, opened 2018-11-28
Feature: ingest and publish articles via REST

lsh-0:
depends on an API-RAML spec before it can be merged.


lsh-0:
```
[52.91.253.82] out: ERROR:xenon:block "src/publisher/api_v2_views.py:134 article_version" has a rank of C
[52.91.253.82] out: use 'radon cc path/to/file.py' for more detail
[52.91.253.82] out: 
```

ha! forgot complexity checking was still in there. I'm glad it's been running this whole time.

lsh-0:
I think the only dubious design in this PR is the token in the POST body to avoid publishing on accidental POSTs. An token in the HTTP header might do the same job. This would be similar to the token that bot-lax proxies between elife-bot and lax

giorgiosironi:
Is https://github.com/elifesciences/api-raml/pull/234/files un-wipped? I'd start from that

lsh-0:
> un-wipped? I'd start from that

no, I need to take a look at why it's failing. And also why the latest commit is failing when it was just linting (sigh)

lsh-0:
closing this PR. See the ticket for more details: https://github.com/elifesciences/issues/issues/4507

PR #417, opened 2018-12-04
Feature: improve identity checks for regular usage of lax

lsh-0:
* removes the mess of unused 'ENV'ironments
* shifts the saving of the new fragment data into `set_article_json` so it has access to the current data (if any) so the identity check can take into account pubdates
* deprecates use of `VALIDATE_FAILS_FORCE`, which, if `False` (default True) will save an empty string as an article's ajson

### DIFF
--- a/src/core/settings.py
+++ b/src/core/settings.py
@@ -50,13 +50,6 @@ SECRET_KEY = cfg('general.secret-key')
 DEBUG = cfg('general.debug')
 assert isinstance(DEBUG, bool), "'debug' must be either True or False as a boolean, not %r" % (DEBUG, )
 
-DEV, VAGRANT, TEST, PROD = 'dev', 'vagrant', 'test', 'prod'
-ENV = cfg('general.env', DEV)
-
-# temporary until the 'vagrant' env is more formalised
-if os.path.exists('/vagrant'):
-    ENV = VAGRANT
-
 ALLOWED_HOSTS = cfg('general.allowed-hosts', '').split(',')
 
 # Application definition
@@ -318,7 +311,7 @@ LOG_NAME = '%s.log' % PROJECT_NAME # ll: lax.log
 
 INGESTION_LOG_NAME = 'ingestion-%s.log' % PROJECT_NAME
 
-LOG_DIR = PROJECT_DIR if ENV == DEV else '/var/log/'
+LOG_DIR = PROJECT_DIR if DEBUG else '/var/log/'
 LOG_FILE = join(LOG_DIR, LOG_NAME) # ll: /var/log/lax.log
 INGESTION_LOG_FILE = join(LOG_DIR, INGESTION_LOG_NAME) # ll: /var/log/lax.log
 
--- a/src/core/settings.py
+++ b/src/core/settings.py
@@ -296,9 +296,11 @@ ENABLE_RELATIONS = True
 # when ingesting an article, if an article says it's related to an article that doesn't exist, should an Article stub be created? default, True.
 RELATED_ARTICLE_STUBS = cfg('general.related-article-stubs', True)
 
+# DEPRECATED: failure to validate article-json should always fail an ingest/publish
 # when ingesting and publishing article-json with the force=True parameter,
 # should validation failures cause the ingest/publish action to fail?
-VALIDATE_FAILS_FORCE = cfg('general.validate-fails-force', True)
+#VALIDATE_FAILS_FORCE = cfg('general.validate-fails-force', True)
+VALIDATE_FAILS_FORCE = True
 
 #
 API_V12_TRANSFORMS = True
--- a/src/publisher/ajson_ingestor.py
+++ b/src/publisher/ajson_ingestor.py
@@ -102,10 +102,6 @@ def _ingest(data, force=False) -> models.ArticleVersion:
     try:
         av, created, updated, previous_article_versions = _ingest_objects(data, create, update, force, log_context)
 
-        # only update the fragment if this article version has *not* been published *or* if force=True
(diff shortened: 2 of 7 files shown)


lsh-0:
Going through this logic revealed some really gnarly twisted old bits that need revisiting.

I'm going to open a separate PR that removes the `VALIDATE_FAILS_FORCE` feature. It goes back to when the article-json and the bot-lax adaptor were still evolving. Nowadays we don't want to ever *knowingly* save invalid article-json fragments (and remove the previously-valid article-json proper). 

I think the `set-article-json` function is getting too large and complex and can be broken down into a simple pipeline of functions. For example:

```python
    pipeline = [
        merge_fragments,
        pre_process,
        validate,
        hashcheck,
        post_process, # removes any temporary things
        save
    ]
```

(which was actually working until I decided to keep this PR as simple as possible).

src/core/settings.py
```diff
@@ -318,7 +313,7 @@ def _load_api_raml(path):
 
 INGESTION_LOG_NAME = 'ingestion-%s.log' % PROJECT_NAME
 
-LOG_DIR = PROJECT_DIR if ENV == DEV else '/var/log/'
+LOG_DIR = PROJECT_DIR if DEBUG else '/var/log/'
```
giorgiosironi: this looks like a separate flag, which is probably just the log path; suppose I turned on debug in end2end to get more introspection about exceptions, etc; I won't expect logs to change folder?

src/publisher/ajson_ingestor.py
```diff
@@ -163,7 +159,10 @@ def _ingest(data, force=False) -> models.ArticleVersion:
         # identity (this data already exists) and validity (this data is malformed)
 
         # validation and hash check of article-json occurs here
-        fragments.set_article_json(av, quiet=False if settings.VALIDATE_FAILS_FORCE else force, hash_check=True)
+        quiet = False if settings.VALIDATE_FAILS_FORCE else force
```
giorgiosironi: since `settings.VALIDATE_FAILS_FORCE` is always `True` at the moment, should `quiet` be simplified to `False`?

src/publisher/ajson_ingestor.py
```diff
@@ -260,7 +259,8 @@ def _publish(msid, version, force=False) -> models.ArticleVersion:
         # merge the fragments we have available and make them available for serving.
         # allow errors when the publish operation is being forced.
         # NOTE: hash checks will always fail on publish events as we modify the `versionDate`.
-        fragments.set_article_json(av, quiet=False if settings.VALIDATE_FAILS_FORCE else force, hash_check=False)
+        quiet = False if settings.VALIDATE_FAILS_FORCE else force
```
giorgiosironi: same simplification may apply here

src/publisher/ajson_ingestor.py
giorgiosironi: but maybe tests patch it to something else, hence this should stay

src/core/settings.py
lsh-0: you wouldn't ever turn DEBUG on in end2end. it's meant purely for development

lsh-0: I'm not sure DEBUG gives you more introspection about exceptions either

src/publisher/ajson_ingestor.py
lsh-0: > but maybe tests patch it to something else

I encountered some of that. Same with update_fragment and I didn't want to be modifying the tests for this unless I absolutely had to.

src/publisher/fragment_logic.py
```diff
@@ -196,21 +199,59 @@ def __init__(self, msg, av, hashval):
...
+        # silent corrections (forced ingest) where only pubdate has changed now requires this
+
+        if oldhash == newhash:
+            # article data is identical
+            # compare pubdates
+            # `preprocess` will alter the publication date value if it hasn't been published yet
+            oldpubdate = raw_original.get('published')
+            newpubdate = result.get('-published')
+
+            if oldpubdate == newpubdate:
```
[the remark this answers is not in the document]
lsh-0: I have the sneaking suspicion that Sian used to be handling all these manually. I showed her how to use the admin interface and I never heard about it again. Production would know. I'll add an `info` for now but it's seems a little half-hearted to the metrics we could be collecting from these apps

lsh-0:
added logging.

src/core/settings.py
giorgiosironi: The question then is what `DEBUG` actually does then (adding more routes?)
I suggest to continue the move from environment name to flag, and introduce the `log_path` configuration in `app.cfg`; a default of `/var/log` seems the safest for rollout, though a default of `.` will make it easier to use locally.

lsh-0: DEBUG comes with Django and is used by Django and is typically expected to be there by third party apps. I'm not going to remove it. The ENV stuff was there since the beginning and it's usage has been slowly eroded until it can be entirely removed.

In lax DEBUG is used to include admin content resources so the /admin interface can be rendered properly while running the development server.

> I suggest to continue the move from environment name to flag, and introduce the log_path configuration in app.cfg; a default of /var/log seems the safest for rollout

I disagree, I think my decision here is best.

lsh-0:
> though I don't have much knowledge of the area

as always, you're not expected to be a domain expert, just a fresh set of eyes and a source of nagging doubt. Your review work is appreciated

src/core/settings.py
lsh-0: if I wasn't clear, DEBUG == native, ENV == me, ages ago. The trick with the urls + DEBUG flag is a common one

PR #436, opened 2019-04-24
Feature: switches project_tests.sh to use .test-green.sh rather than .test.sh

lsh-0:
an experiment to see if this brings down the test run time

cc @gnott , @giorgiosironi

### DIFF
--- a/.test-green.sh
+++ b/.test-green.sh
@@ -12,6 +12,12 @@ if [ ! -z "$args" ]; then
     module="src.$args"
 fi
 
+# remove any old compiled python files
+find src/ -name '*.pyc' -delete
+
+# called by test.sh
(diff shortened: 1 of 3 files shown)


lsh-0:
it fails because it doesn't generate the expected artifacts, but the tests run for 27mins vs 45mins.

still quite long though. the test suite takes 1min 50secs locally using sqlite. Using postgresql (longer to setup and teardown) I'd expect maybe 5mins?

gnott:
If all the tests pass then of course please choose the test runner you prefer. Maybe you want to try running the tests again now Giorgio has switched to t2 unlimited.

lsh-0:
for posterity: it was the cpu credits in the end

PR #437, opened 2019-04-29
Feature: replaces autopep8 with Black

lsh-0:
the code changes are probably overwhelming, but the significant changes happened in:
* requirements.txt
* .scrub.sh

https://github.com/ambv/black

cc @giorgiosironi


src/core/middleware.py
```diff
@@ -22,48 +24,50 @@ def process_request(self, request):
         # if request doesn't have all expected headers, strip auth
         if CGROUPS not in request.META:
             # no auth or invalid auth request, return immediately
-            LOG.debug('header %r not found, refusing auth', CGROUPS)
+            LOG.debug("header %r not found, refusing auth", CGROUPS)
```
GiancarloFusiello: did black add double quotes to everything?

lsh-0: yup

lsh-0:
I'm assuming the first paragraph on the Black website was at least glanced at. Important bits for me are:

> By using it, you agree to cede control over minutiae of hand-formatting. In return, Black gives you speed, determinism, and freedom from pycodestyle nagging about formatting.

> Black makes code review faster by producing the smallest diffs possible.

I don't agree with or understand all of the changes it's making, but removing that particular thorn in my side is worth it.

PR #454, opened 2019-07-31
Fix: bumps psycopg2

lsh-0:
it's mysteriously segfaulting on new instances:
https://github.com/elifesciences/lax-formula/pull/36

possibly caused by an upgrade to libpq ...?

fyi @giorgiosironi


giorgiosironi:
Since we are on it, is `psycopg-binary` easier to install? Not sure if the segfault was at installation or runtime.

lsh-0:
it was at runtime. I'll take a look at this -binary package

PR #457, opened 2019-08-05
Feature: replaces 'green' with 'pytest'

lsh-0:
* replaces 'green' test runner with 'pytest' 
* pushes coverage's configuration back into the rcfile

cc @giorgiosironi


lsh-0:
Related to this [PR](https://github.com/elifesciences/lax/pull/281) that fixes some tests accessing the net, pytest has a plugin that lets you disable that behaviour

.test.sh
```diff
@@ -21,7 +21,9 @@ find src/ -name '*.pyc' -delete
 rm -f build/junit.xml
 
 # testing management commands that require a queue shared between processes
-LAX_MULTIPROCESSING=1 coverage run --source='src/' --omit='*/tests/*,*/migrations/*' src/manage.py test "$module" --no-input
+export DJANGO_SETTINGS_MODULE=core.settings
```
lsh-0: require by pytest-django, I can also push this into a pytest.ini type file, but it's fine for now

.test.sh
```diff
@@ -21,7 +21,9 @@ find src/ -name '*.pyc' -delete
...
+export LAX_MULTIPROCESSING=1
+pytest "$module" -vvv --cov=src --cov-config=.coveragerc
```
lsh-0: pytest pathing differs from the default python and green test runners but it's still the first arg

.test.sh
giorgiosironi: Would try to keep a manual `pytest ...` command easy to write, but your call. Could also be in `conftest.py` if there is one.

lsh-0: eh, I'll leave it as-is.

welcome back! I bet you have a zillion emails

PR #494, opened 2020-01-17
django bump

lsh-0:
bumps django, django dependencies and pretty much everything except pylint


src/publisher/tests/test_utils.py
```diff
@@ -118,21 +118,23 @@ def test_json_dumps_rfc3339(self):
         self.assertEqual(utils.json_dumps(struct), expected)
 
     def test_json_dumps_rfc3339_on_non_utc(self):
-        tz = pytz.timezone(
-            "Australia/Adelaide"
-        )  # +9.5 hours ahead, but pytz thinks it's only +9
+        # nice, but don't do this. the timezone changes with DST
```
giorgiosironi: it does; though if you used a fixed input date that is not experiencing DST, the timezone library should convert it accordingly? e.g. Jun 1st rather than Jan 1st?

lsh-0: Yes, that was my understanding too, but for some reason I was getting a 14 minute discrepancy I couldn't bend my head around at that time of day.

And magic dates are not much better than magic numbers

PR #521, opened 2020-06-03
Updated bot-lax-adaptor to: Merge pull request #629 from elifesciences/update_elife_tools/112

elife-alfred-user:
I have run https://alfred.elifesciences.org/job/dependencies-lax-update-bot-lax-adaptor/351/display/redirect which resulted in this pull request.


gnott:
This introduces the structured abstract parsing by using the latest `elifetools` library. @lsh-0 I wanted to check whether you think this is safe to merge and send on its way to being deployed. I expect we'd only get validation errors if an article XML with a `<sec>` tag comes through; regular `<p>` tag abstracts will produce the same results, as demonstrated by the corpus generation being all valid in tests.

lsh-0:
erm, lax does validation as well. It may get generated correctly and pass validation in bot-lax, but lax will reject it as invalid unless it is updated as well.

in fact, where is the lax ticket for the structured abstract work? this structured abstracts thing definitely feels uncoordinated.

gnott:
Yes, there's a bit of a chicken and egg scenario. One bright spot is if we continue receiving regular abstracts in article XML then it is all backwards compatible.

PR #529, opened 2020-06-15
removes 'abstract' from snippet creation

lsh-0:
* removes 'abstract' from the snippet creation

- [x] add test to ensure 'abstract' isn't present in response or in database


lsh-0:
*wow*, these tests are **slow**

lsh-0:
the fact that it passes without errors means I have nothing testing the expected output of an article listing page /facepalm

gnott:
I ran tests for lax for the first time a couple weeks ago and I thought it was just me for why the tests seemed slow, not having installed or configured it properly. They were running for maybe fifteen minutes?

lsh-0:
That figure sounds about right.

Some time in the past I switched the test runner from Green to pytest and I suspect it was Green running the tests in parallel and pytest running them serially.

lsh-0:
here we go. I fully expect some end2end tests to fail

lsh-0:
huh no errors. well, good.

PR #531, opened 2020-06-19
structured abstracts bump

lsh-0:
ticket: https://github.com/elifesciences/issues/issues/5693

* removes vor/poa v1<->v2 transformations
* adds support for poa v3 and poa v4
* begun slow transition to pytest-style tests where possible. Tests derived from the Django base class will setup a database and do migrations etc. It's part of the reason the tests have become so slow.

- [x] review


src/core/settings.py
```diff
@@ -190,12 +190,13 @@ def cfg(path, default=0xDEADBEEF):
     "DEFAULT_PERMISSION_CLASSES": [],
     # 'DEFAULT_CONTENT_NEGOTIATION_CLASS': 'publisher.negotiation.eLifeContentNegotiation',
     "DEFAULT_RENDERER_CLASSES": (
+        # order is important, most to least specific
         "publisher.negotiation.ArticleListVersion1",
+        "publisher.negotiation.POAArticleVersion3",  # experimental, structured abstracts
```
lsh-0, line 195: rm 'experimental', it won't be for much longer

src/core/settings.py
```diff
@@ -190,12 +190,13 @@ def cfg(path, default=0xDEADBEEF):
...
         "publisher.negotiation.POAArticleVersion2",
-        "publisher.negotiation.POAArticleVersion1",
+        "publisher.negotiation.VORArticleVersion4",  # experimental, structured abstracts
```
lsh-0, line 197: rm 'experimental'

src/core/settings.py
```diff
@@ -228,14 +224,16 @@ def cfg(path, default=0xDEADBEEF):
...
 SCHEMA_PATH = join(PROJECT_DIR, "schema/api-raml/dist")
 
-# a response is valid if validates under any version of it's schema.
+# a response is valid if it validates under any version of it's schema.
 # order is important. if all attempts to validate fail, the first validation error is re-raised
 ALL_SCHEMA_IDX = {
     "poa": [
+        (3, join(SCHEMA_PATH, "model/article-poa.v3.json")),
         (2, join(SCHEMA_PATH, "model/article-poa.v2.json")),
         (1, join(SCHEMA_PATH, "model/article-poa.v1.json")),
```
lsh-0, line 233: rm poa v1

src/core/settings.py
```diff
@@ -228,14 +224,16 @@ def cfg(path, default=0xDEADBEEF):
...
     ],
     "vor": [
+        (4, join(SCHEMA_PATH, "model/article-vor.v4.json")),
         (3, join(SCHEMA_PATH, "model/article-vor.v3.json")),
         (2, join(SCHEMA_PATH, "model/article-vor.v2.json")),
         (1, join(SCHEMA_PATH, "model/article-vor.v1.json")),
```
lsh-0, line 239: rm vor v1

src/publisher/middleware.py
```diff
@@ -315,38 +212,128 @@ def middleware(request):
...
+        client_accepts_poa_versions = [
+            int(row[-1]) for row in client_accepts_poa_list if isint(row[-1])
+        ]
+
+        if not client_accepts_poa_versions:
+            # no specific version specified, return latest version
+            return response
+
+        max_accepted_poa = max(client_accepts_poa_versions)
+        current_poa_version = 3  # TODO: fix
```
lsh-0, line 311: see settings/ALL_SCHEMA_IDX

src/publisher/api_v2_views.py
```diff
@@ -21,8 +21,8 @@
 
 def ctype(status):
     return {
-        POA: "application/vnd.elife.article-poa+json; version=2",
-        VOR: "application/vnd.elife.article-vor+json; version=3",
+        POA: "application/vnd.elife.article-poa+json; version=3",
```
lsh-0, line 24: see settings/ALL_SCHEMA_IDX

src/publisher/middleware.py
```diff
@@ -143,12 +56,12 @@ def requested_version(request, response):
     return (response_mime[0], "*" if not versions else max(versions))
 
 
-def deprecated(request):
-    if not settings.API_V12_TRANSFORMS:
-        return False
+def is_deprecated(request):
     targets = [
         ("application/vnd.elife.article-poa+json", "version", b"1"),
```
lsh-0, line 61: rm. poa v1 has been removed

src/publisher/middleware.py
```diff
@@ -143,12 +56,12 @@ def requested_version(request, response):
...
         ("application/vnd.elife.article-vor+json", "version", b"1"),
```
lsh-0, line 62: rm vor v1 has been removed

src/publisher/middleware.py
```diff
@@ -143,12 +56,12 @@ def requested_version(request, response):
...
+        # 2020-06
+        ("application/vnd.elife.article-vor+json", "version", b"2"),
```
lsh-0, line 64: can we say the oldest in the group is deprecated? in which case, see settings/ALL_SCHEMA_IDX

src/publisher/middleware.py
```diff
@@ -161,45 +74,12 @@ def deprecated(request):
...
-
-
-def apiv1_deprecated(get_response_fn):
+def deprecated(get_response_fn):
     def middleware(request):
         response = get_response_fn(request)
-        if deprecated(request):
-            response["warning"] = "Deprecation: Support for version 1 will be removed"
+        if is_deprecated(request):
+            msg = "Deprecation: Support for this version will be removed"
```
lsh-0, line 81: "support for this content-type version will be removed"

src/publisher/middleware.py
```diff
@@ -228,6 +108,8 @@ def middleware(request):
 
 
 def content_check(get_response_fn):
+    "compares content-type requests against what is supported."
```
lsh-0, line 111: "compares content-type in request against ..."

src/publisher/middleware.py
```diff
@@ -302,7 +197,9 @@ def v3_vor_valid_under_v2(ajson):
     return True
 
 
-def incompatible_v2_check(get_response_fn):
+def downgrade_vor_content_type(get_response_fn):
+    """if a content-type less than the current VOR version is requested, downgrade content-type if possible or return a 406"""
```
lsh-0, line 201: break on 'or return'

src/publisher/middleware.py
```diff
@@ -315,38 +212,128 @@ def middleware(request):
...
+            return response
+
+        max_accepted_vor = max(client_accepts_vor_versions)
+        body = json.loads(response.content.decode("utf-8"))
+
+        if max_accepted_vor == 3:
+            # client specifically accepts a v3 VOR only
+            # we might be ok if the content is valid under v3
+            if vor_valid_under_v3(body):
+                # all good, drop content-type returned to VOR v2
```
lsh-0, line 240: VOR v3

src/publisher/middleware.py
```diff
@@ -315,38 +212,128 @@ def middleware(request):
...
+                new_response = HttpResponse(
+                    response.content, content_type=new_content_type
+                )
+                # this is where RESTResponses keep it
+                new_response.content_type = new_content_type
+                return new_response
+
+        if max_accepted_vor == 2:
+            # client specifically accepts a v2 VOR only (deprecated)
+            # we might be ok if the content is valid under v3 and v2
```
lsh-0, line 252: elaborate on 'v3 and v2'

src/publisher/middleware.py
```diff
@@ -315,38 +212,128 @@ def middleware(request):
...
+                new_content_type = "application/vnd.elife.article-vor+json; version=2"
+                new_response = HttpResponse(
+                    response.content, content_type=new_content_type
                 )
+                # this is where RESTResponses keep it
+                new_response.content_type = new_content_type
+                return new_response
 
-        return response
+        # nuts, we have a v4-only article and client had a very limited Accept header
```
lsh-0, line 264: rm. logic has been changed, new comment required

src/publisher/middleware.py
```diff
@@ -315,38 +212,128 @@ def middleware(request):
...
+
+    def middleware(request):
+        request_accept_header = request.META.get("HTTP_ACCEPT", "*/*")
+        response = get_response_fn(request)
+
+        if response.status_code != 200:
+            # unsuccessful response, ignore
+            return response
+
+        # are we returning VOR content?
```
lsh-0, line 288: POA

src/publisher/middleware.py
```diff
@@ -315,38 +212,128 @@ def middleware(request):
...
+        ]
+
+        if not client_accepts_poa_versions:
+            # no specific version specified, return latest version
+            return response
+
+        max_accepted_poa = max(client_accepts_poa_versions)
+        current_poa_version = 3  # TODO: fix
+
+        if max_accepted_poa == current_poa_version:
```
lsh-0, line 313: do I do this with VORs?

src/publisher/middleware.py
```diff
@@ -315,38 +212,128 @@ def middleware(request):
...
+            # user requested the current latest version of
+            return response
+
+        body = json.loads(response.content.decode("utf-8"))
```
lsh-0, line 317: rm

src/publisher/middleware.py
lsh-0: no, I wasn't...

lsh-0:
hey @gnott , I've added you as a reviewer but don't feel pressed to understand what all the changes are, just let me know if you spot anything egregious.

code formatting is handled exclusively by the [black](https://github.com/psf/black) library so I won't be changing anything it would revert on it's next run.

src/publisher/api_v2_views.py
```diff
@@ -18,12 +18,15 @@
...
+CURRENT_POA_VERSION = str(settings.ALL_SCHEMA_IDX["poa"][0][0])
+CURRENT_VOR_VERSION = str(settings.ALL_SCHEMA_IDX["vor"][0][0])
+
 
 def ctype(status):
     return {
-        POA: "application/vnd.elife.article-poa+json; version=2",
-        VOR: "application/vnd.elife.article-vor+json; version=3",
-        ERR: "application/json",  # text/plain as well?
+        POA: "application/vnd.elife.article-poa+json; version=" + CURRENT_POA_VERSION,
```
gnott, line 27: I assume string conversion of `CURRENT_POA_VERSION` is absolutely safe here if it's an `int`?

src/publisher/middleware.py
```diff
@@ -76,61 +32,19 @@ def flatten_accept(header, just_elife=False):
     return lst
 
 
+def has_structured_abstract(ajson):
+    "returns True if ajson contains a structured abstract"
+    if "abstract" in ajson:
+        # a regular abstract is just the keys 'type' and 'text'
+        # a structured abstract also has an 'id' field
+        return "id" in ajson["abstract"]["content"][0]
```
gnott, line 40: It is possible (which I think was the plan) that the `id` is an optional attribute for an abstract section I requested, so if the API schema is still that way, then a structured abstract may not have any `id` values.

src/publisher/middleware.py
```diff
@@ -143,15 +57,25 @@ def requested_version(request, response):
...
-    targets = [
-        ("application/vnd.elife.article-poa+json", "version", b"1"),
-        ("application/vnd.elife.article-vor+json", "version", b"1"),
-    ]
+# deprecated content-types are simply the oldest ones
+
+DEPRECATED_VOR = (
+    "application/vnd.elife.article-vor+json",
+    "version",
+    bytes(str(settings.ALL_SCHEMA_IDX["vor"][-1][0]), "utf-8"),
```
gnott, line 65: A smart way to deprecate the oldest with `-1` 👍

src/publisher/tests/test_api_v2_views.py
```diff
@@ -18,6 +18,172 @@
...
+    msid = 31549  # vor spec version 4, our first
+    fixture_path = join(
+        base.FIXTURE_DIR, "structured-abstracts", "elife-31549-v1.xml.json"
+    )
+    fixture = json.load(open(fixture_path, "r"))["article"]
+    mock = Mock(article_json_v1=fixture, status="vor")
+    vor_v3_ctype = "application/vnd.elife.article-vor+json; version=3"
+
+    with patch("publisher.logic.article_version", return_value=mock):
+        url = reverse("v2:article-version", kwargs={"msid": msid, "version": 1})
```
gnott, line 145: I cannot exactly see how the test fixture data is version 4, does it relate to the `v2` or `1` in this line? If not, please ignore my comment, but I don't know what these values signify.

src/publisher/tests/test_api_v2_views.py
```diff
@@ -251,6 +307,7 @@ def test_article_list(self):
         "a list of published articles are returned to an unauthenticated response"
         resp = self.c.get(reverse("v2:article-list"))
         self.assertEqual(resp.status_code, 200)
+        # TODO: fix
```
gnott, line 310: Are still todo?

src/publisher/api_v2_views.py
lsh-0: strings can be concatenated with `+` in python.

`CURRENT_POA_VERSION` is defined just above this section of code and is wrapped in a call to `str`.

Should that `str` ever be removed any test calling a view (that in turn calls `ctype`) would fail immediately.

src/publisher/middleware.py
lsh-0: nice catch. I'll take a second look at the structured abstract spec

src/publisher/middleware.py
lsh-0: yeah ... I found I was hardcoding versions all over the place in the code. Back when it was just a change from v1 to v2 it wasn't such a big deal and the bump from v2 to v3 was trivial but now it's getting tiresome ;)

I wonder if the RAML tooling for python has changed much since when we first adopted it. It was brand new back then without much adoption.

src/publisher/tests/test_api_v2_views.py
lsh-0: yes, it's confusing.

this fixture: `/path/to/fixture-dir/structured-abstracts/elife-31549-v1.xml.json` is valid against VOR spec version 4 only - it contains a structured abstract and will fail validation in spec versions 1 to 3.

If a user requests in their 'accepts' header a spec version other than 4 (`3` in this case), we won't be able to satisfy it and would expect a HTTP 406 response.

This `"v2:article-version"` refers to the API version within Lax, because before the elife 2.0 infrastructure, lax had a different API with a different set of paths serving up a different set of article data (EIF, remember that?). When 2.0 came along I need to support both the old urls and new urls while everybody transitioned thus the new set of urls got the `v2` [namespace](https://docs.djangoproject.com/en/3.0/topics/http/urls/#url-namespaces).

The `article_json_v1` is an unfortunate naming choice of mine for a database field when we were adding the 2.0 support. The idea was that as new spec versions of the article json were added we would add another field to the article model to hold the article json valid for that version. This hasn't happened and is unlikely to happen. That field should be renamed.

lsh-0: oh, and this `kwargs={"msid": msid, "version": 1}` is a parameter to the `/articles/$msid/version/$version` endpoint. In this context it's not being helpful ;) I'll remove it.

src/publisher/tests/test_api_v2_views.py
lsh-0: /sigh 

yes, I suppose I should or it won't ever get done.

src/publisher/tests/test_api_v2_views.py
lsh-0: fyi @NuclearRedeye 

The above is an example of how gnarly programs can get over time without proper maintenance. Imagine being new to the lax project and having to understand all that? Yet how does one prioritise and justify the time for a task like "rename db column" or "remove v1 api" against focused sprints? It goes into a stack, gets prioritised against everything else, sinks to the bottom, is politely taken off the board and gets forgotten. There is no epic for this. 

And this developer that is new to the project will come in with their task to do, become overwhelmed and (probably) advocate that it be torn down and rebuilt. In fact, this is exactly what happened with [elife-dashboard](https://github.com/elifesciences/elife-dashboard) and why we have a stillborn [elife-dashboard-2](https://github.com/elifesciences/elife-dashboard/tree/develop/dashboard_2).

And we see this pattern happening again with Kubernetes, github actions and Prometheus dashboards.

src/publisher/tests/test_api_v2_views.py
```diff
@@ -18,6 +18,172 @@
...
+        "application/vnd.elife.article-poa+json; version=1",
+        # 2, cannot fulfill request (fixture is a poa, not a specific obsolete vor)
+        "application/vnd.elife.article-vor+json; version=1",
+        # 3, cannot fulfill request (fixture is a poa, not a specific valid vor)
+        "application/vnd.elife.article-vor+json; version=4",
+        # 4, cannot fulfill request (fixture is a poa, not a general vor)
+        "application/vnd.elife.article-vor+json",
+        # 5, cannot fulfill request (fixture is still a poa)
+        "application/vnd.elife.article-vor+json; version=1, application/vnd.elife.article-vor+json; version=2",
+        # 6, fictious content versions (for now)
```
lsh-0, line 113: spelling.

src/publisher/middleware.py
lsh-0: [here it is in our spec](https://github.com/elifesciences/api-raml/commit/93bda43a661c71c19468a936880793363a4205bd#diff-5095a45df81a60f9049fd5ce3545a678R8-R26)

It's not 100% clear to me, but it reads as if: an abstract is either ('`oneOf`') a paragraph type or a map with the `type`, `id`, `title` and `content` keys ('`properties`'). There is no `required` section listing the required properties and none of the properties have `required: True` against them and  [the default value for `required` appears to be `True`](https://github.com/raml-org/raml-spec/blob/master/versions/raml-10/raml-10.md#property-declarations).

So we should be safe with just checking for an `id` value. @gnott , I hate to be a pain but could you chase down where you heard that `id` should be optional? It might have been a requirement that was overlooked, or a requirement that was dropped or I've completely misread the spec (cc @nlisgo )

src/publisher/tests/test_api_v2_views.py
lsh-0: @gnott I've removed the endpoint being patched so there are fewer mentions of 'version' floating around confusing matters.

src/publisher/middleware.py
gnott: Here I think is my first comment/request for adding an optional `@id` attribute, however in the end after the changes to the schema, I didn't check whether it is still optional https://github.com/elifesciences/api-raml/pull/246#discussion_r434232366

Whether the `@id` attribute will always be there for eLife articles would be something to run by Production and the typesetter I think, but conceivably a structured abstract could omit any `@id` from the `<sec>` tags and be acceptable.

lsh-0: it looks like a requirement that was overlooked then (if I'm not misreading the spec). @nlisgo to please confirm.

lsh-0: after our chat, I'll switch this to look for `type` rather than `id`. I expect to merge this Thursday and then the api-raml changes made after this PR makes it into production

src/publisher/tests/test_api_v2_views.py
gnott: > EIF, remember that?

I've forgotten it sometimes, but now you mention it, I remember. The EIF support was removed on the `elife-bot` side of things, but it was kicking around for a while beyond where it was actually in use.

Your extra details here cleared up much of my confusion, and if you simplified it a little, all the better for understanding.

src/publisher/middleware.py
lsh-0: update: turns out 'type' is shared with regular abstracts as well. I've changed the test so that a structured abstract is simply missing the 'text' key.

lsh-0:
https://alfred.elifesciences.org/job/test-lax/

lsh-0:
argh! what a bastard. I've just noticed it refused to push my change to the `id` check: https://github.com/elifesciences/lax/pull/531/files#diff-575bc49e0226016bef2ffb50f108b41bR40

I'll have to follow up with another PR. damn.

lsh-0:
here: https://github.com/elifesciences/lax/pull/534

PR #537, opened 2020-07-03
lax, vor v2 removal

lsh-0:
Removes support for VOR v2 content type.

We're only going to support two versions of an article type at a time, so POA v2 and v3 and VOR v3 and v4.

- [x] review


lsh-0:
I've pinned the api-raml to the vor removal branch to see where lax depends on hardcoded values.

new-content-types.md
```diff
@@ -0,0 +1,10 @@
+# adding new content-type versions to lax
+
+- [ ] start a new branch
+- [ ] ensure the api-raml has these new specs present
+- [ ] open src/core/settings.py for editing
+- [ ] change REST_FRAMEWORK, add new content type version
+- [ ] change ALL_SCHEMA_IDX, add new content type version
+- [ ] in `src/publisher/negotiation.py`, change `_dynamic_types` and add new content type version
+- [ ] in `src/publisher/middleware.py`, add any content-type downgrades
+
```
gnott, line 10: This looks handy as a reminder for next time, I know personally after a number of months I forget this amount of detail 👍

lsh-0: I often keep checklists like this around. Whether I remember to look here next time is another thing. It also illustrates that there are too many things that need to be updated. I navigate a lot of these older projects by instinct and muscle memory now ;)

PR #607, opened 2021-02-26
removes support for deprecated content types and content 'downgrading'

lsh-0:


lsh-0:
yeah, this isn't ever going to happen.

PR #611, opened 2021-03-02
reverts bot-lax pin to generate v5 content

lsh-0:
reverts commit 70d3839935f8070704ea265bc503de846bce9879

fyi @gnott , we'll create another PR for the bot-lax bump


gnott:
We can revert the revert for the PR to reinstate it

PR #619, opened 2021-03-10
Try ping in smoke tests before articles endpoint.

gnott:
A theory if we try the ping endpoint in smoke tests prior to trying to get the article list it may be easier to determine when a `502` Bad Gateway response is returned in a test pipeline after the entire stack has been restarted.

### DIFF
--- a/smoke_tests.sh
+++ b/smoke_tests.sh
@@ -1,13 +1,17 @@
 #!/bin/bash
 set -ex
 
-for path in / /api/v2/ping /api/v2/articles; do
-    [ $(curl --write-out %{http_code} --silent --output /dev/null https://$(hostname)$path) == 200 ]
+num_attempts=3
+retry_delay=5 # seconds
+
+for path in / /api/v2/articles; do
+    [ $(curl --retry $num_attempts --retry-delay $retry_delay --write-out %{http_code} --silent --output /dev/null https://$(hostname)$path) == 200 ]
 done
 
 for path in /admin /explorer; do
-    [ $(curl --write-out %{http_code} --silent --output /dev/null https://$(hostname)$path) == 301 ]
+    [ $(curl --retry $num_attempts --retry-delay $retry_delay --write-out %{http_code} --silent --output /dev/null https://$(hostname)$path) == 301 ]
 done
+
 
 # api-raml version must be identical for lax and bot-lax-adaptor
 diff /opt/bot-lax-adaptor/api-raml.sha1 api-raml.sha1
--- a/smoke_tests.sh
+++ b/smoke_tests.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 set -ex
 
-for path in / /api/v2/articles; do
+for path in / /api/v2/ping /api/v2/articles; do
     [ $(curl --write-out %{http_code} --silent --output /dev/null https://$(hostname)$path) == 200 ]
 done
 


lsh-0:
thanks for opening a ticket. The change looks fine but while the smoke tests are under the microscope I want to switch over to the smoke library and add some retry logic too:

* https://github.com/elifesciences/builder-base-formula/blob/939ce52f7f2ddd4ff8e9b56eedde481100ad4c46/elife/smoke.sls
* https://github.com/elifesciences/journal/blob/b602f0f994543427b35ea4a547220dca648d5ecf/smoke_tests.sh

gnott:
I'm less sure this small change will help to diagnose any further, because `/` may be served by Django, and if so, then the article list is the one exceeding 15 second timeout. At least, that's all I was able to figure out by looking at it and making some guesses. We can close this PR if you want to do something else.

lsh-0:
I'm modifying it now

lsh-0:
the `smoke_url_ok` and `retry` links I posted turned out to be less than useful. curl has some retry and delay capabilities built in these days, they may not have been in there ubuntu 16.04 times

lsh-0:
> is the one exceeding 15 second timeout

where is this 15 second timeout you mentioned? is it a default in curl or something in Jenkins?

gnott:
I got the 15 value from https://github.com/elifesciences/lax-formula/blob/master/salt/lax/config/etc-nginx-sitesavailable-lax.conf#L61, which seeemed to match the log datetimes, 15 seconds between the GET and the status code 502 returned.

lsh-0:
hrm, I may have to investigate why they are that way. The harakiri value in uwsgi.ini is 79 seconds (!) which indicates that 15s value once used to be 80s.

lsh-0:
anyway, I'll merge and we'll see what happens

gnott:
There's another which is 80s https://github.com/elifesciences/lax-formula/blob/master/salt/lax/config/etc-nginx-sitesenabled-bot-lax-adaptor.conf#L59 ... yeah

lsh-0:
makes a bit more sense for bot-lax, because its webapp is actually converting content to article-json and validating it, where lax isn't doing any such thing.

lsh-0:
```
15:10:39  [54.85.117.153] run: smoke-tests /srv/lax
15:10:39  [54.85.117.153] out: + num_attempts=3
15:10:39  [54.85.117.153] out: + retry_delay=5
15:10:39  [54.85.117.153] out: + for path in / /api/v2/articles
15:10:39  [54.85.117.153] out: +++ hostname
15:10:39  [54.85.117.153] out: ++ curl --retry 3 --retry-delay 5 --write-out '%{http_code}' --silent --output /dev/null https://prod--lax.elifesciences.org/
15:10:39  [54.85.117.153] out: + '[' 200 == 200 ']'
15:10:39  [54.85.117.153] out: + for path in / /api/v2/articles
15:10:39  [54.85.117.153] out: +++ hostname
15:10:39  [54.85.117.153] out: ++ curl --retry 3 --retry-delay 5 --write-out '%{http_code}' --silent --output /dev/null https://prod--lax.elifesciences.org/api/v2/articles
15:10:39  [54.85.117.153] out: + '[' == 200 ']'
15:10:39  [54.85.117.153] out: ./smoke_tests.sh: line 8: [: ==: unary operator expected
15:10:39  [54.85.117.153] out: Exit code of /usr/local/bin/smoke-tests: 2
15:10:39  [54.85.117.153] out: 
```

just shoot me

lsh-0:
I can replicate it:

```
elife@continuumtest--lax:/srv/lax$ /bin/bash -c "sudo systemctl restart nginx &" && ./smoke_tests.sh
+ num_attempts=3
+ retry_delay=5
+ for path in / /api/v2/articles
+++ hostname
++ curl --retry 3 --retry-delay 5 --write-out '%{http_code}' --silent --output /dev/null https://continuumtest--lax.elifesciences.org/
+ '[' 000 == 200 ']'
```

gnott:
I wonder if the the request is attempted again after 5 seconds, but the original request still hasn't reached timeout, then there would be no http_code returned. How about retry delay of 20 seconds, exceeds the 15 second timeout?

lsh-0:
new PR: https://github.com/elifesciences/lax/pull/621

lsh-0:
> 15:10:39  [54.85.117.153] out: ++ curl --retry 3 --retry-delay 5 --write-out '%{http_code}' --silent --output /dev/null https://prod--lax.elifesciences.org/api/v2/articles
15:10:39  [54.85.117.153] out: + '[' == 200 ']'

No, in the original error something happened so curl didn't even run. Because it didn't send a request it got no response to print using it's `--write-out` param. It also sends all output to /dev/null so - who knows?

lsh-0:
If half of one's life is spent sleeping, then half of the remainder is spent waiting for CI to finish building something. Would love to get the corpus generation and lax CI times down to half of what they are

gnott:
Was not the request to `/` successful, therefore the port 80 is up?  `20:40:39  [34.234.82.201] out: + '[' 200 == 200 ']'`

lsh-0:
> edit: actually, that's not quite true.

new PR addresses one error I was seeing where the app server was running but nginx was in the middle of rebooting and curl was returning 000

The command where curl failed to provide any feedback at all is a mystery. If you can recreate it let me know

gnott:
I tested a little more, I also get `000` from curl if the response is `502`.

The `--retry` flag seems to be suboptimal, because even when it receives a `200` response, curl still tries it three times anyway, with the delay in between. It might be better to restructure the retry loop to not depend on curl, instead to check the `http_code` value on each attempt, and if `200` then continue to the next GET.

NuclearRedeye:
> The --retry flag seems to be suboptimal, because even when it receives a 200 response, curl still tries it three times anyway

Odd, the man page for curl implies that it should only retry for http 5xx errors...

```
--retry <num>
If a transient error is returned when curl tries to perform a transfer, it will retry this number of times before giving up. Setting the number to 0 makes curl do no retries (which is the default). Transient error means either: a timeout, an FTP 5xx response code or an HTTP 5xx response code.
When curl is about to retry a transfer, it will first wait one second and then for all forthcoming retries it will double the waiting time until it reaches 10 minutes which then will be the delay between the rest of the retries. By using --retry-delay you disable this exponential backoff algorithm. See also --retry-max-time to limit the total time allowed for retries. (Added in 7.12.3)

If this option is used multiple times, the last occurrence decide the amount.
```

Could that be because of our use of the `--retry-delay` flag, might be worth trying `--retry-max-time` instead?

PR #672, opened 2021-08-03
removes some unused/easily replaceable dependencies

lsh-0:


src/publisher/utils.py
```diff
@@ -378,9 +379,13 @@ def unique(seq):
 #
 
 
-@cache
 def load_schema(schema_path):
```
lsh-0, line 382: perhaps eagerly load all schemas in settings.py rather than this lazy loading

src/publisher/management/commands/rehash.py
```diff
@@ -15,23 +13,20 @@ class Command(BaseCommand):
...
+                num = q.count()
+                # `.iterator()` to use server-side cursor and avoid thrashing memory
+                # https://docs.djangoproject.com/en/3.2/ref/models/querysets/#iterator
+                for i, av in enumerate(q.iterator()):
                     av.article_json_hash = fragment_logic.hash_ajson(av.article_json_v1)
                     av.save()
-                    t.update(1)
-                t.info("done")
+                    # 13 of 152
+                    print("%s or %s" % (i, num))
```
lsh-0, line 27: "of"

PR #756, opened 2022-05-10
updates modcommand.py to be compatible with django 3.2

lsh-0:


lsh-0:
should fix the current problem in end2end

